PR #2, opened 2021-02-26
Add basic CI infrastructure.

### LINKED ISSUE #1: Add CI to this repository
Should tests some version of the following:

- Install Zeek from binary package
- Install zkg (unless 4.0, where is comes with)
- Configure zkg (autoconf / `@load packages`)
- `zkg install .` from a clone of this repository
- Check that our analyzers show up in `zeek -NN _Zeek::Spicy` output (ideally create a baseline showing all the included analyzers)
- Check that running `zeek` uses the new analyzers. maybe run `zeek` and then baseline `cat loaded_scripts.log  | grep spicy-analyzers`; our analyzer scripts should show up in there.

Do this for a couple different platforms and (for now) Zeek 3.0, Zeek 3.2, Zeek 4.0.

bbannier:
Closes #1.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,24 @@
+package_ci_task:
+  timeout_in: 120m
+  container:
+    dockerfile: ci/Dockerfile
+    matrix:
+      # TODO(bbannier): Update this once packages for zeek-4.0.0 are available as zeek-lts.
+      - docker_arguments:
+        - ZEEK_LTS: 1
+        - ZEEK_VERSION: 3.0.13-0
+      - docker_arguments:
+        - ZEEK_LTS:
+        - ZEEK_VERSION: 3.2.4-0
+      - docker_arguments:
+        - ZEEK_LTS:
+        - ZEEK_VERSION: 4.0.0-0
+    cpu: 4
+    memory: 12G
+
+  install_package_script:
+    - zkg install --force .
+
+  check_install_script:
+    - zeek -NN _Zeek::Spicy | grep Analyzer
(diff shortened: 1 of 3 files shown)


bbannier:
Tests for builds outside of the Spicy tree fail since d8837eec456a600cffa341dba76883806bd60b81. This needs to be fixed before we can make progress here.

rsmmr:
> Tests for builds outside of the Spicy tree fail since [d8837ee](https://github.com/zeek/spicy-analyzers/commit/d8837eec456a600cffa341dba76883806bd60b81). This needs to be fixed before we can make progress here.

What are you seeing? Just tried, it's working for me (at least after the recent merge of the `analyzers` branch, but I believe I had tried it before that too).

rsmmr:
CI passing now, assigning back for finalizing/clean up.

.cirrus.yml
```diff
@@ -0,0 +1,21 @@
...
+  container:
+    dockerfile: ci/Dockerfile
+    matrix:
+      - docker_arguments:
+        - ZEEK_LTS: 1
+        - ZEEK_VERSION: 3.0.13-0
+      - docker_arguments:
+        - ZEEK_LTS:
+        - ZEEK_VERSION: 3.2.4-0
+      # TODO(bbannier): add zeek-4.0 LTS release.
```
bbannier, line 12: The packages of the non-LTS version will switch to 4.0.0 later today; LTS will stay at 3.0.x for another 2 weeks or so.

Let's update the non-LTS to point to the latest later today.

rsmmr (approved):

PR #5, opened 2021-03-04
PNG analyzer

0xxon:


analyzer/file/png/PNG.spicy
```diff
@@ -0,0 +1,194 @@
...
+#import Exif;
+
+public type File = unit {
+	signature: bytes &size=8; # has to be 137 80 78 71 13 10 26 10
+	var colour_type: uint8 = 0;
+	: Chunk(self)[] &eod;
+
+	on signature {
+		local d: spicy::Base64Stream;
+		local thesig : bytes = spicy::base64_decode(d, b"iVBORw0KGgo=");
```
rsmmr: Is this to get binary data in there? You can do `b"\xYY"`, would that work here? And they maybe just with a `&requires` attribute for the field?

analyzer/file/png/main.zeek
```diff
@@ -0,0 +1,106 @@
...
+
+redef record fa_file += {
+	png: Info &optional;
+};
+
+const png_mime_types = { "image/png" };
+
+event zeek_init() &priority=5
+	{
+	#Files::register_for_mime_types(Analyzer::get_tag("SPICY_PNG"), png_mime_types);
```
rsmmr, line 62: So this (or some of this) is still not working?

rsmmr (commented):
This is great, nice work!

analyzer/file/png/main.zeek
0xxon: It's not necessary - the evt file has the mime type. So - I should just remove this, sorry.

analyzer/file/png/PNG.spicy
0xxon: This is no longer in there - I changed this to your suggestion.

PR #8, opened 2021-03-04
Load wireguard

0xxon:


rsmmr:
@0xxon looks like a Zeek version issue in `main`: `counter` used to be reserved word.

0xxon:
Ups, let me fix this.

PR #9, opened 2021-03-08
Fix counter.

keithjjones:
@0xxon I ran into an error with this while playing with Zeek v3.0.2 and think it's associated with your change here:  https://github.com/zeek/spicy-analyzers/commit/bc89e749fcc84c3d009aae406e0e12dd009f2645#diff-7d486216965ddcd85079f5fd0c03a94913754c9e0624a750978585b389c39900


0xxon:
yup, I missed the test.

I wonder why CI did not pick that up...

0xxon (approved):

PR #14, opened 2021-03-11
Make wireguard a bit less likely to have false positives

### LINKED ISSUE #13: Wireguard false positives
I believe I'm seeing Wireguard false positives running this on a large university network (sorry I can't post their data).  I think it's because protocolconfirmation is being called for every parsed type, so it's not seeing the full handshake before calling the first protocolconfirmation:  

https://github.com/zeek/spicy-analyzers/blob/main/analyzer/protocol/wireguard/wireguard_zeek.spicy#L9

I'm seeing traffic identified as wireguard where there are a ton of packets in one direction but zero in the other direction, according to conn.log.  This is likely a false positive.  I'm making this issue to track work on fixing this.  I think I can fix it, but I'm currently learning Spicy so it might not be right away.

0xxon:
A few small changes to the protocol to check that we are not accidentally parsing other packets.

Might be enough to close #13

### DIFF
--- a/analyzer/protocol/wireguard/dpd.sig
+++ b/analyzer/protocol/wireguard/dpd.sig
@@ -2,6 +2,6 @@
 
 signature wireguard_packet {
   ip-proto == udp
-  payload /^(\x01|\x02|\x03|\x04)\x00\x00\x00/
+  payload /^(\x01|\x02|\x03)\x00\x00\x00/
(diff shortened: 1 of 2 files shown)


0xxon:
Looks like CI is broken?

rsmmr (approved):

Issue #10, opened 2021-03-09
DNS event arguments error


keithjjones:
On Zeek v4.0.0 I get this error with the current master in cluster mode:

```
#separator \x09
#set_separator  ,
#empty_field    (empty)
#unset_field    -
#path   reporter
#open   2021-03-09-16-41-42
#fields ts      level   message location
#types  time    enum    string  string
1615326102.956359       Reporter::ERROR Event parameter mismatch, expected 6 parameters, but got 5      /Users/keith.jones/.zkg/clones/package/spicy-analyzers/analyzer/protocol/dns/dns.evt, line 17
1615326121.433034       Reporter::ERROR Event parameter mismatch, expected 6 parameters, but got 5      /Users/keith.jones/.zkg/clones/package/spicy-analyzers/analyzer/protocol/dns/dns.evt, line 17
1615326121.434854       Reporter::ERROR Event parameter mismatch, expected 6 parameters, but got 5      /Users/keith.jones/.zkg/clones/package/spicy-analyzers/analyzer/protocol/dns/dns.evt, line 17
1615326124.872783       Reporter::ERROR Event parameter mismatch, expected 6 parameters, but got 5      /Users/keith.jones/.zkg/clones/package/spicy-analyzers/analyzer/protocol/dns/dns.evt, line 17
```
I get similar errors with other versions of Zeek and other protocols (other than wireguard because it's new) in this repo.  I assume one would have to disable the original DNS analyzer to use the new DNS events?  I thought the new dns events would be in a different namespace, but it's causing a conflict with the original ones.  Thanks!

keithjjones:
Thanks for the reply @rsmmr.  

I've installed this in two places (three total installs) with similar results.  1) On a macos macbook with Big Sur. That's the log you see above.  I took zeek v4.0, compiled and installed in cluster mode.  I then compiled and installed spicy.  I then zkg installed spicy-analyzers and I get this error in this issue after I deploy the cluster.  It errors on its own on dns traffic, I don't have to send anything special through.

In scenario 2 I have this installed on 2 identical Linux appliances with Zeek v3.0.2. I compiled spicy on the appliance with all the latest flex+bison+cmake+gcc and I get this same error.  The appliance also says there are the wrong number of arguments and seems to have the original analyzer events loaded as well.

In all cases when I comment out in the cmake file the new spicy analyzers that conflict with the original ones in Zeek, the problems go away.  

[...]

PR #16, opened 2021-03-15
[DNS] Add missing event parameters.

rsmmr:
Zeek >= 3.2 has added the original query to some of its DNS events. This
adapts the Spicy-generated events accordingly.

Addresses #10.


bbannier (dismissed):

tests/protocol/dns/basic.zeek
```diff
@@ -3,7 +3,15 @@
 # @TEST-EXEC: ${ZEEK} -r ${TRACES}/dns53.pcap %INPUT
 # @TEST-EXEC: btest-diff conn.log
 # @TEST-EXEC: btest-diff dns.log
+# @TEST-EXEC: if zeek-version 32000; then btest-diff .stdout; fi
```
bbannier, line 6: Hmm, did you intend to perform this check _exclusively for 3.2_?  All the other checks here are for >=3.2 (assuming below `Version::number >= 32000` is equivalent to `ZEEK_VERSION >= 30200` in the evt file).

Maybe we could alternatively hardcode the expected output for versions < 3.2.

rsmmr: The script checks for `>=`  on its end if it's used like that. But turns out I forgot to commit the script. It's the same as in `spicy-plugin`.

bbannier: Ah, I see 🤦

bbannier (dismissed):

PR #18, opened 2021-03-15
Update README.

rsmmr:


README.md
```diff
@@ -7,9 +7,9 @@ through [Spicy](https://github.com/zeek/spicy).
 
 Currently, the following analyzers are included:
 
-- DHCP
-- DNS
-- HTTP
+- DHCP [1]
+- DNS [1]
+- HTTP [1]
```
bbannier, line 12: Since GH markdown doesn't support footnotes and the references here aren't clickable, maybe we should just add the footnote text in parentheses, e.g.,

```suggestion
- DHCP (replaces corresponding Zeek analyzer)
- DNS (replaces corresponding Zeek analyzer)
- HTTP (replaces corresponding Zeek analyzer)
```

README.md
```diff
@@ -91,6 +93,27 @@ Now `zeek -NN _Zeek::Spicy` should show similar output as above.
 When you run Zeek, add `spicy-analyzers` to the command line to load
 the analyzer scripts.
 
+## Configuration
+
+By default, all included analyzers will be activated, they will
```
bbannier, line 98:
```suggestion
By default, all included analyzers will be activated, and they will
```

README.md
```diff
@@ -91,6 +93,27 @@ Now `zeek -NN _Zeek::Spicy` should show similar output as above.
...
+want to disable one of the Spicy analyzers, you can do so by calling
+one of the built-in functions
+[disable_protocol_analyzer/disable_file_analyzer()](https://docs.zeek.org/projects/spicy/en/latest/zeek.html#functions).
+For example, to disable the HTTP analyzer, add this to your
+`site.zeek`:
+
+    event zeek_init()
+        {
+        Spicy::disable_protocol_analyzer(Analyzer::ANALYZER_SPICY_HTTP);
+        }
```
bbannier, line 109: Not sure if that applies here, but one can select a syntax for code blocks which would then be used e.g., when GH renders these markdown docs in its webui,

```suggestion
```.zeek
event zeek_init()
    {
    Spicy::disable_protocol_analyzer(Analyzer::ANALYZER_SPICY_HTTP);
    }
```

README.md
```diff
@@ -91,6 +93,27 @@ Now `zeek -NN _Zeek::Spicy` should show similar output as above.
...
+
+You can find the `ANALYZER_*` value to use for an analyzer in the
+output of `zeek -N _Zeek::Spicy`.
+
+(Note that `disable_file_analyzer()` requires a current development
```
bbannier, line 114: Can we reference a SHA here?

bbannier (approved):

README.md
rsmmr: Had it like that first, but am preferring the pseudo-footnote for brevity; have another one coming too

README.md
rsmmr: Not worth it I think, people following Zeek master should keep updating anyways, and not stick to an old version.

README.md
bbannier: In above suggestion I forgot the terminating line, i.e., add the following to the end

    ```

bbannier (approved):

PR #19, opened 2021-03-17
Extend PNG test to catch Spicy's #817.

rsmmr:
Note: turns out the file analyzer tests currently don't run because they aren't listed in the btest config. Wil fix that separately.


bbannier (commented):
Why does this test pass before we have merged zeek/spicy#838? 😅

rsmmr:
Because the test isn't active, see other PR

bbannier (approved):

PR #26, opened 2021-03-18
Install Spicy from upstream package.

bbannier:


ci/Dockerfile
```diff
@@ -3,17 +3,12 @@
 
 FROM ubuntu:focal
 
-# Increase this to force a rebuild of the Docker image (in Cirrus, in particular).
```
rsmmr, line 6: Same comments here for the plugin: Suggest to leave it in.

rsmmr (dismissed):

PR #29, opened 2021-03-24
Reenable PE analyzer tests.

bbannier:


bbannier:
At least on my machine this still has issues as the printing of `set` values is not stable. Maybe we need to fix this in Spicy proper (probably: for `map`s as well).

tests/Scripts/canonify-zeek-log
```diff
@@ -12,6 +12,6 @@ fi
 # File IDs changed between Zeek 3.1 and 3.2.
 
 ${sed} 's/^ *#(open|close).(19|20)..-..-..-..-..-..$/#\1 XXXX-XX-XX-XX-XX-XX/g' | \
-    ${sed} 's/F[A-Za-z0-9]{16,17}/XXXXXXXXXXXXXXXXX/g' | \
+    ${sed} 's/F[A-Za-z0-9]{15,17}/XXXXXXXXXXXXXXXXX/g' | \
```
rsmmr, line 15: Is this needed? I don't see baseline change going along with this.

rsmmr (approved):

tests/Scripts/canonify-zeek-log
bbannier: Yes, I believe it is as on zeek-3.0.13 I see a `fuid` of `FOuWFKf04xcHH4ck` emitted in the log.

There is not baseline change all strings get normalized to the same string `XXXXXXXXXXXXXXXXX` which is already in the existing baseline.

PR #30, opened 2021-03-24
Address TODO in OpenVPN analyzer.

### LINKED ISSUE #27: Investigate TODO in OpenVPN analyzer
There's a TODO in the OpenVPN Spicy code:

https://github.com/zeek/spicy-analyzers/blob/358904848246358097d90ece6bb2929bc1fa3098/analyzer/protocol/openvpn/openvpn.spicy#L61

This would shorten the code quite a bit, not clear why it isn't working, need to investigate.

bbannier:
The reason the commented out code was not working was an off-by-one
error as the size of the `OpenVPNRecord` needs to include not only the
opcode-specific payload, but also the message type.

Closes #27.


keithjjones (approved):
Nice catch!

rsmmr (approved):
Indeed a nice catch!

PR #33, opened 2021-03-25
Wireguard: change logging to only have one line per connection

0xxon:
This commit revamps the wireguard logging. Now there is only one line
per connection that tracks connection establishment as well as the
number of handshake initiation and response packets.

Addresses GH-13

### DIFF
--- a/analyzer/protocol/wireguard/main.zeek
+++ b/analyzer/protocol/wireguard/main.zeek
@@ -43,8 +43,10 @@ export {
 	## Event raised for the wireguard packet_data packet
 	global wireguard::packet_data: event(c: connection, is_orig: bool, receiver_index: count, key_counter: count, encapsulated_packet_length: count);
 
+@ifdef ( Conn::register_removal_hook )
 	## Wireguard finalization hook; wireguard information is logged in it
 	global finalize_wireguard: Conn::RemovalHook;
+@endif
 }
 
 redef record connection += {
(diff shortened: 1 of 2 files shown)


0xxon:
ci is broken again? :)

rsmmr:
> ci is broken again? :)

it's this: https://github.com/zeek/spicy/issues/865

I just kicked off the tasks again.

rsmmr:
Looks like we have a couple of legitimate CI failures now:

- looks like the change doesn't work with Zeek 3.0
- @bbannier there are more PE diffs here, looks like only some lines were canonified?

bbannier:
Regarding the PE diffs, I accidentally merged an outdated branches which does not pass CI. I fixed this with d8a5ac21d0c45bce49a2acaa38405c1b8e8dc8d7 and rebasing this PR onto at least that version or merging it should remove the error seen here.

0xxon:
Ok, in perhaps stupid questions - since we don't actually officially support Zeek 3.0 anymore - is there a reason why we want this repository to be compatible with it?

0xxon:
standalone CI seems to be broken - and I don't think that's my fault :)

bbannier:
The failure you see for _standalone ci_ is zeek/spicy-plugin#27.

0xxon:
Should to work in Zeek 3 + 4 now.

rsmmr:
Please try a rebase to main and see if that fixes the CI issues now.

Re: 3.0: Technically, we're still in that 2 month window, but the main reason is really that there hasn't been a specific reason to break with 3.0 yet, given that it's all working fine and nothing has come up that would have made it tricky to support for now.

0xxon:
Yup, that worked :)

rsmmr (approved):

PR #34, opened 2021-03-26
Install scripts as well in CMake install

rsmmr:
Also let package depend on zeek-plugin.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -35,6 +35,7 @@ package_ci_task:
     - zkg install --force --skiptests .
 
   check_script:
+    - zeek -NN _Zeek::Spicy | grep -q ANALYZER_SPICY_TFTP
     - zeek -NN _Zeek::Spicy | grep Analyzer
     - (cd /tmp && zeek -r ${CIRRUS_WORKING_DIR}/tests/Traces/tftp_rrq.pcap local && test -e tftp.log && grep rfc1350.txt tftp.log)
 
@@ -70,3 +71,4 @@ standalone_ci_task:
   check_script:
     - zeek -NN _Zeek::Spicy | grep Analyzer
     - PATH=/opt/spicy-plugin/bin:$PATH make -C tests test-install
+    - (cd /tmp && zeek -r ${CIRRUS_WORKING_DIR}/tests/Traces/tftp_rrq.pcap spicy-analyzers && test -e tftp.log && grep rfc1350.txt tftp.log)
--- a/analyzer/CMakeLists.txt
+++ b/analyzer/CMakeLists.txt
@@ -3,3 +3,15 @@
 add_subdirectory(file)
 add_subdirectory(protocol)
 
+if ( SPICY_SCRIPTS_OUTPUT_DIR_INSTALL )
+    if ( SPICY_IN_TREE_BUILD )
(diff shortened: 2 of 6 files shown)


analyzer/CMakeLists.txt
```diff
@@ -3,3 +3,15 @@
 add_subdirectory(file)
 add_subdirectory(protocol)
 
+if ( SPICY_SCRIPTS_OUTPUT_DIR_INSTALL )
+    if ( SPICY_IN_TREE_BUILD )
+        # Do not install by default but tie to install-spicy-analyzers target.
+        install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/ DESTINATION "${SPICY_SCRIPTS_OUTPUT_DIR_INSTALL}/spicy-analyzers/"
```
bbannier, line 9: `SPICY_SCRIPTS_OUTPUT_DIR_INSTALL` is constructed from `CMAKE_INSTALL_FULL_LIBDIR` which is an absolute path so this might cause issues down the line (e.g., in `DESTDIR` builds).

Can we use a relative path here? Maybe it would be possible to initialize `SPICY_SCRIPTS_OUTPUT_DIR_INSTALL` from `CMAKE_INSTALL_LIBDIR` instead?

tests/Scripts/get-path
```diff
@@ -28,7 +28,7 @@ elif [ "$1" = "--plugin-path-for-installation" ]; then
     # 3. If we have spicyz in path, derive from module path.
     # TODO: Could add seperate --print-plugin-dir to spicyz.
     spicyz=$(which spicyz 2>/dev/null)
-    test -n "${spicyz}" && echo $(cd $(${spicyz} --print-module-path)/../_Zeek_Spicy && pwd) && exit 0
+    test -n "${spicyz}" && echo $(cd $(${spicyz} --print-module-path)/../../_Zeek_Spicy && pwd) && exit 0
```
bbannier, line 31:
```suggestion
    test -n "${spicyz}" && (cd "$(${spicyz} --print-module-path)"/../../_Zeek_Spicy && pwd) && exit 0↲
```

analyzer/CMakeLists.txt
rsmmr: `SPICY_SCRIPTS_OUTPUT_DIR_INSTALL` is actually usually coming out of `spicyz --print-scripts-path` (i.e., from external). The `CMAKE_INSTALL_FULL_LIBDIR` version is actually just a fallback, not even sure we still need that (but didn't want to remove as part of this in case it would break something else).

.cirrus.yml
```diff
@@ -70,3 +71,4 @@ standalone_ci_task:
   check_script:
     - zeek -NN _Zeek::Spicy | grep Analyzer
     - PATH=/opt/spicy-plugin/bin:$PATH make -C tests test-install
+    - (cd /tmp && zeek -r ${CIRRUS_WORKING_DIR}/tests/Traces/tftp_rrq.pcap spicy-analyzers && test -e tftp.log && grep rfc1350.txt tftp.log)
```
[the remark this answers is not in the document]
rsmmr: I think the likelihood is very small, and we're aren't doing that anywhere else either.

PR #38, opened 2021-03-26
Fix in-tree Spicy build.

rsmmr:
Thanks, CMake.


bbannier (approved):
:shipit: if this passes CI, but I have no idea why CMake makes this behave differently between the two versions.

rsmmr:
Sorry, should have posted the CMake error message that the old version produced:

```
CMake Error at zeek/spicy-analyzers/analyzer/CMakeLists.txt:9 (install):
  install DIRECTORY does not allow "COMPONENT" after PATTERN or REGEX.
```

PR #39, opened 2021-03-30
Document the sources of our traces files.

rsmmr:
Still need to add the URLs for a couple of OpenVPN traces.


keithjjones:
https://wiki.wireshark.org/OpenVPN

rsmmr:
> https://wiki.wireshark.org/OpenVPN

I found two of the traces there, did I miss the other two?

keithjjones:
I got them from a researcher that I'm pretty sure came from this bug report:  https://wireshark-bugs.wireshark.narkive.com/IwG4sfj5/bug-8240-new-dissector-for-the-openvpn-protocol

rsmmr:
I don't find the traces themselves on that page, but I take it you are positive that these are ok to redistribute?

tests/Traces/README
```diff
@@ -1,13 +1,32 @@
...
+- [dhcp.pcap](https://gitlab.com/wireshark/wireshark/-/wikis/uploads/__moin_import__/attachments/SampleCaptures/dhcp.pcap)
 
+DNS
+- [dns53.pcap](https://github.com/zeek/zeek/blob/master/testing/btest/Traces/dns53.pcap)
 
+HTTP
+- [http-post.pcap](https://github.com/zeek/zeek/blob/master/testing/btest/Traces/http/post.trace)
+
+OpenVPN
+- [openvpn-tcp-tls-auth.pcap](https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=OpenVPN_TCP_tls-auth.pcapng)
```
bbannier, line 16: This link seems to be deprecated and by banner redirects to https://wiki.wireshark.org/SampleCaptures, but the file is listed on neither of the overview pages. I am not sure what that means for us.

tests/Traces/README
```diff
@@ -1,13 +1,32 @@
...
+- [openvpn.pcap](https://wireshark-bugs.wireshark.narkive.com/IwG4sfj5/bug-8240-new-dissector-for-the-openvpn-protocol)
+- [openvpn_tcp_nontlsauth.pcap](https://wireshark-bugs.wireshark.narkive.com/IwG4sfj5/bug-8240-new-dissector-for-the-openvpn-protocol)
```
bbannier, line 18: It would be nice to point to the specific attachment while we still have some context as there seem to be multiple versions. Searching for _Created attachment_ in the bug report yields links to the different attachments.

tests/Traces/README
rsmmr: Yeah, I didn't find the trace at the new location. That old link still leads directly to download for me, that's why I just left it in as is.

tests/Traces/README
rsmmr: Nice, I hadn't see the attachment links. So I found `openvpn_tcp_nontlsauth.pcap` now, but not `openvpn.pcap`; none of the attachment to that ticket seems to match the latter.

bbannier: Hmm, not ideal, but looks like we cannot do much here now.

PR #48, opened 2021-04-14
Fix indents and remove wrapper.

keithjjones:
Fixing a few things I saw.

### DIFF
--- a/analyzer/protocol/ipsec/main.zeek
+++ b/analyzer/protocol/ipsec/main.zeek
@@ -533,7 +533,7 @@ export {
 	## is_orig: True if the message was sent by the originator.
 	##
 	## msg: The parsed IPSec message.
-    global ipsec::ikev1_sa_payload: event(c: connection, is_orig: bool, msg: ipsec::IKEv1_SA_Msg);
+	global ipsec::ikev1_sa_payload: event(c: connection, is_orig: bool, msg: ipsec::IKEv1_SA_Msg);
 
 	## Fires on every IKEv1 vendor ID payload.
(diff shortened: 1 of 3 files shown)


bbannier (commented):
While you are at it, would you mind fixing the mixed indents in the blocks starting at l.737 and l.742 as well?

keithjjones:
Got it, thanks!

analyzer/protocol/ipsec/main.zeek
```diff
@@ -672,8 +672,8 @@ function set_session(c: connection)
 event protocol_confirmation(c: connection, atype: Analyzer::Tag, aid: count) &priority=5
 	{
 	if ( atype == Analyzer::ANALYZER_SPICY_IPSEC_IKE_UDP ||
-	     atype == Analyzer::ANALYZER_SPICY_IPSEC_TCP ||
-	     atype == Analyzer::ANALYZER_SPICY_IPSEC_UDP )
+		 atype == Analyzer::ANALYZER_SPICY_IPSEC_TCP ||
```
bbannier, line 675: FWIW, I think this we'd actually like to keep: we indent this line with a tab like the `if` on the line above, and then do manual aligning with spaces. Indenting with tabs here will give inconsistent results depending on a user's tabwith.

bbannier (changes requested):

analyzer/protocol/ipsec/main.zeek
keithjjones: Switched to spaces.

PR #55, opened 2021-05-04
Initial version of a ZIP analyzer.

rsmmr:
This decompresses ZIP files and feeds their content back recursively into Zeek's file analysis.

It comes with the following limitations:

    - We support only deflated content for uncompression.

    - We do not yet support ZIP64 extensions (but report them)

    - ZIP doesn't lend itself well to stream parsing, so we have to use
      some heuristics. In particular, we process all file content even
      if the central directory doesn't list it. (Could be considered a
      plus for Zeek actually!).

    - We heuristically scan for the end of file content where the size
      it not known upfront. The heuristic performs a CRC check, which
      should make it quite robust.


analyzer/file/zip/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+
+
+if ( SPICY_VERSION_NUMBER GREATER_EQUAL 10100 )
```
bbannier, line 3:
```suggestion
if ( SPICY_VERSION_NUMBER GREATER_EQUAL 10100 )
```

analyzer/file/zip/zeek_zip.spicy
```diff
@@ -0,0 +1,33 @@
+
+module Zeek_ZIP;
```
bbannier, line 2:
```suggestion
module Zeek_ZIP;
```

analyzer/file/zip/main.zeek
```diff
@@ -0,0 +1,44 @@
+module ZIP;
+
+export {
+	type File: record {
+		## True if from global directory header, false if from local file header
+		global_: bool;
+		## File ID associated with content analysis of this file. Only available for local
+		## headers where file content has been further processed.
+		fid: string;
```
bbannier, line 9: Can we mark this as `&optional` or does this not work?

analyzer/file/zip/zip.spicy
```diff
@@ -0,0 +1,205 @@
+#
+# TODO:
```
bbannier, line 2:
```suggestion
# TODO:
```

analyzer/file/zip/zip.spicy
```diff
@@ -0,0 +1,205 @@
...
+    flags:              bitfield(16) {
+        encrypted: 0 &convert=cast<bool>($$);
+        compression: 1..2;
+        has_descriptor: 3;
+        enhanced_deflation: 4;
+        compressed_patched: 5;
+        strong_encryption: 6;
+        efs: 11;
+        masked_headers: 13;
+    };
```
bbannier, line 108: Since this seems identical tothe `flags` in `DirectoryHeader`, it would be nice to move the RHS into a field declaration, and reuse it here and in `DirectoryHeader`.

tests/file/zip/descriptor.zeek
```diff
@@ -0,0 +1,23 @@
...
+# @TEST-EXEC: btest-diff files.log
+# @TEST-EXEC: btest-diff .stdout
+# @TEST-EXEC: btest-diff extracted.log
+#
+# @TEST-DOC: Test ZIP analyzer with a download of a ZIP file that uses a data descriptor.
+#
+# With the data descriptor, the file sizes comes only after the file content. Our
+# ZIP archive stress-tests the heuristic for finding the end of the file content by
+# containing a file that includes the terminator magic string inside its plain-text
+# content.
```
bbannier, line 13: 👍

bbannier (approved):

analyzer/file/zip/zip.spicy
rsmmr: Good point, will factor this out.

analyzer/file/zip/main.zeek
rsmmr: done

bbannier (approved):

bbannier (commented):
I just realized that the trace added here triggers virus scanners (by design). Could we replace the file so that e.g., users don't have to fight their virus scanners putting that file into quarantine?

rsmmr:
> I just realized that the trace added here triggers virus scanners (by design). Could we replace the file so that e.g., users don't have to fight their virus scanners putting that file into quarantine?

Ack, swapped it out.

bbannier (approved):

PR #56, opened 2021-05-11
LDAP analyzer

mmguero:
The LDAP analyzer (and its underlying ASN.1 module) is to the point where it has the things I need from it. I've tested it on as much live traffic and PCAP traffic as I could and it seems to be working well.

Here's what it has:

- ASN.1 structure decoding: this is probably generally useful for more than just the LDAP parser, so it may be of interest for this to be included somehow as part of spicy's standard modules or whatever
  - everything is working except for the "constructed" forms of `ASN1BitString` and `ASN1OctetString`
- LDAP: the LDAP parsing is basically "done once" through a single call to `ASN1Message` (which parses itself recursively) and then the application-level data is also parsed via `&parse-from` a byte array belonging to the outer ASN.1 sequence. This second level of parsing is also done using the ASN.1 data types.
  - events
    - `ldap::message` - called for each LDAP message
    - `ldap::bindreq` - when a bind request is made
    - `ldap::searchreq` - basic search request information
    - `ldap::searchres` - called each time a search result is returned
  - enums
    - `ProtocolOpcode`
    - `ResultCode`
    - `BindAuthType`
    - `SearchScope`
    - `SearchDerefAlias`
    - `FilterType`
  - Zeek log files
    - `ldap.log` - contains information about all LDAP messages except those that are search-related. Log lines are grouped by connection ID + message ID
      - `ts` (time)
      - `uid` (connection UID)
      - `id` (connection ID 4-tuple)
      - `proto` (transport protocol)
      - `message_id` (LDAP message ID)
      - `version` (LDAP version for bind requests)
      - `opcode` (set of 1..n operations from this uid+message_id)
      - `result` (set of 1..n results from this uid+message_id)
      - `diagnostic_message` (vector of 0..n diagnostic message strings)
      - `object` (vector of 0..n "objects," the meaning of which depends on the operation)
      - `argument` (vector of 0..n "argument," the meaning of which depends on the operation)
    - `ldap_search.log` - contains information about LDAP searches. Log lines are grouped by connection ID + message ID
      - `ts` (time)
      - `uid` (connection UID)
      - `id` (connection ID 4-tuple)
      - `proto` (transport protocol)
      - `message_id` (LDAP message ID)
      - `scope` (set of 1..n search scopes defined in this uid+message_id)
      - `deref` (set of 1..n search deref alias options defined in this uid+message_id)
      - `base_object` (vector of 0..n search base objects specified)
      - `result_count` (number of result entries returned)
      - `result` (set of 1..n results from this uid+message_id)
      - `diagnostic_message` (vector of 0..n diagnostic message strings)
  - test
    - basic tests for detecting plugin presence and simple bind and search result/requests

Here's what it doesn't have, which could be added by future parties interested in expanding it:

- although LDAP can use UDP as transport, currently the analyzer only looks for `389/tcp, 3268/tcp`; this may be easy to change, but I don't have any examples of traffic to test it with so I haven't bothered
- LDAP [referrals](https://tools.ietf.org/html/rfc4511#section-4.1.10) are not parsed out of the results
- [SASL credentials](https://datatracker.ietf.org/doc/html/rfc4511#section-4.2) in bind requests are not being parsed beyond the mechanism string
- SASL information in bind responses are not being parsed; for that matter, SASL-based LDAP stuff hasn't been tested much and may have issues
- Search filters and attributes: while basic parsing is done in the `AttributeSelection`, `AttributeValueAssertion`, `SubstringFilter` and `SearchFilter` units, I'm not really pulling stuff out from the search filter tree (as the protocols allows arbitrarily complex combinations of AND, OR, substring, etc. that I don't think would be easily representable in a log file)
- the details of `SearchResultReference` are not being parsed
- the only detail of `ModifyRequest` being parsed is the object name
- the details of `AddRequest` are not being parsed
- the details of `ModDNRequest` are not being parsed
- the details of `CompareRequest` are not being parsed
- the details of `AbandonRequest` are not being parsed
- the details of `ExtendedRequest` are not being parsed
- the details of `ExtendedResponse` are not being parsed
- the details of `IntermediateResponse` are not being parsed


analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+      } else if (self?.bitstring) {
+        print "%.*s%s %s %s" % ((depth*2)+1, DEBUG_SPACES, head.tag.class, head.tag.tpe, "BITSTRING"), self.bitstring, head.tag.constructed;
+      } else if (self?.octetstring) {
+        print "%.*s%s %s %s" % ((depth*2)+1, DEBUG_SPACES, head.tag.class, head.tag.tpe, "OCTETSTRING"), self.octetstring, head.tag.constructed;
+      } else if (self?.objectidentifier) {
+        print "%.*s%s %s %s" % ((depth*2)+1, DEBUG_SPACES, head.tag.class, head.tag.tpe, "OBJECTIDENTIFIER"), self.objectidentifier;
+      } else if (self?.asn1string) {
+        print "%.*s%s %s %s" % ((depth*2)+1, DEBUG_SPACES, head.tag.class, head.tag.tpe, "STRING"), self.asn1string;
+      } else if (self?.seq) {
+        print "%.*s%s %s %s" % ((depth*2)+1, DEBUG_SPACES, head.tag.class, head.tag.tpe, "SEQUENCE|SET"), |self.seq.submessages|, head.len.len;
```
bbannier, line 270: This combination of stringification, tuples, and`|...|` triggers the CI error you are seeing.

We have the PR zeek/spicy#919 open to fix this; alternatively and as a temporary workaround you could pull the size computation into a variable and print that,i.e.,

```.r
local num_submessages = |self.seq.submessages|;
print ... % ... num_submessages ...;
```

If you decide to not wait, but go with a workaround, leaving a `TODO` referencing the PR would be great.

mmguero: I put the workaround in as you suggested, with a TODO referencing the PR. Thanks. 7737a0c

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+
+};
+
+
+#- ASN.1 message header (tag + length information) ----------------------------
+
+public type ASN1Header = unit {
+  tag: ASN1Tag;
+
+  var total_length: uint64;
```
bbannier, line 204: This seems unused. Let's drop it.

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+#- ASN.1 message body ---------------------------------------------------------
+
+public type ASN1Body = unit(head: ASN1Header, recursive: bool, depth: uint32) {
+  switch ( head.tag.tpe ) {
+
+    ASN1Type::BOOLEAN -> bool_value: uint8 &convert=cast<bool>($$) {
+      if (head.len.len != 1) {
+        throw "ASN1Type::BOOLEAN detected with length != 1";
+      }
+    }
```
bbannier, line 220: If your check expresses a protocol invariant (i.e., we'd only ever hit it for invalid input), let's instead express it with a `&requires` attribute, e.g.,
```suggestion
    ASN1Type::BOOLEAN -> bool_value: uint8 &convert=cast<bool>($$) &requires=head.len.len==1;
    }
```

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+  }
+};
+
+#- ASN.1 message (headless) ---------------------------------------------------
+# same rules as ASN1Message for body/application_data
+
+public type ASN1MessageHeadless = unit(head: ASN1Header, recursive: bool, depth: uint32) {
+  var depth: uint32 = depth;
```
bbannier, line 317: Same question as on the previous unit, why the copy?

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+        print "%.*s%s %d %s %d" % ((depth*2)+1, DEBUG_SPACES, self.head.tag.class, self.application_id, self.head.tag.constructed, self.head.len.len), self.head;
+      }
+    }
+  }
+};
+
+#- ASN.1 message (headless) ---------------------------------------------------
+# same rules as ASN1Message for body/application_data
+
+public type ASN1MessageHeadless = unit(head: ASN1Header, recursive: bool, depth: uint32) {
```
bbannier, line 316: This seems unused right now. Is that due to this PR not implementing all of ASN.1 or LDAP?

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+
+#- ASN.1 message helper units for recursive (set/sequence) vs. not ------------
+
+public type ASN1MessageBaseRecursive = unit {
+  msg : ASN1Message(True, 0);
+};
+
+public type ASN1MessageBase = unit {
+  msg : ASN1Message(False, 0);
+};
```
bbannier, line 343: These seem unused.

bbannier (commented):
Thanks for this PR. I made a first round through the ASN.1 parser. Will look at the other parts now.

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+    ASN1Type::BITSTRING -> bitstring: ASN1BitString(head.len.len, head.tag.constructed);
+
+    ASN1Type::OCTETSTRING -> octetstring: ASN1OctetString(head.len.len, head.tag.constructed)
+                                          &convert=$$.value.decode(hilti::Charset::ASCII);
+
+    ASN1Type::OBJECTIDENTIFIER -> objectidentifier: ASN1ObjectIdentifier(head.len.len)
+                                                    &convert=$$.oidstring;
+
+    ASN1Type::PrintableString,
+    ASN1Type::GeneralizedTime,
```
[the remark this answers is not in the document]
mmguero: [GeneralizedTime](https://www.obj-sys.com/asn1tutorial/node14.html) is represented as [a string value](https://en.wikipedia.org/wiki/GeneralizedTime) as far as I can tell, so I think this is correct.

You're right about the other string types missing here, I think. From what I can gather on wikipedia and RFCs the other string types you mentioned are also encoded this way. I'll change the list to this:

```
    ASN1Type::BMPString,
    ASN1Type::CHARACTERSTRING,
    ASN1Type::GeneralizedTime,
    ASN1Type::GeneralString,
    ASN1Type::GraphicString,
    ASN1Type::IA5String,
    ASN1Type::NumericString,
    ASN1Type::PrintableString,
    ASN1Type::TeletextString,
    ASN1Type::UTCTime,
    ASN1Type::UTF8String,
    ASN1Type::VideotextString,
    ASN1Type::VisibleString,
    ASN1Type::UniversalString -> asn1string: ASN1String(head.tag, head.len.len)
                                             &convert=$$.value;
```

I will be honest some of the things here were migrated from [rsmmr/hilti](https://github.com/rsmmr/hilti/blob/master/libspicy/parsers/asn1.spicy#L145-L158) and just made the same assumptions made there. Thanks for the thorough review, I'll continue to go through the rest of them.

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+    ASN1Type::UTCTime,
+    ASN1Type::UTF8String,
+    ASN1Type::BMPString,
+    ASN1Type::UniversalString -> asn1string: ASN1String(head.tag, head.len.len)
+                                             &convert=$$.value;
```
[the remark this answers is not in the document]
mmguero: Excellent point, thank you.

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+  UNIVERSAL = 0,
+  APPLICATION = 1,
+  CONTEXT_SPECIFIC = 2,
+  PRIVATE = 3
+};
+
+#- ASN.1 tag definition (including length) ------------------------------------
+
+type LengthType = unit {
+
```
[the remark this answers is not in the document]
mmguero: assuming this is part of the next comment

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+public type ASN1Class = enum {
+  UNIVERSAL = 0,
+  APPLICATION = 1,
+  CONTEXT_SPECIFIC = 2,
+  PRIVATE = 3
+};
+
+#- ASN.1 tag definition (including length) ------------------------------------
+
+type LengthType = unit {
```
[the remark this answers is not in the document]
mmguero: In this case we can't because sometimes we need `.len` and sometimes we need `.tag_len`.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  NOT_SET        = 255
+};
+
+type AttributeSelection = unit(depth: uint32) {
+  var attributes: vector<string>;
+
+  # TODO: parse AttributeSelection as per
+  # https://tools.ietf.org/html/rfc4511#section-4.5.1
+  # and decide how deep that should be fleshed out.
+  seq: asn1::ASN1Message(True, depth) {
```
bbannier, line 301: Candidate for hiding?

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+      }
+    }
+  }
+};
+
+type AttributeValueAssertion = unit(depth: uint32) {
+  var desc: string = "";
+  var val: string = "";
+
+  seq: asn1::ASN1Message(True, depth) {
```
bbannier, line 318: Candidate for hiding?

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+    #}
+  }
+
+};
+
+type SearchFilter = unit(depth: uint32) {
+  var filterType: FilterType = FilterType::NOT_SET;
+  var filterBytes: bytes = b"";
+
+  filter : asn1::ASN1Message(True, depth) {
```
bbannier, line 360: Candidate for hiding?

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+                                             &default=SearchDerefAlias::NOT_SET;
+  sizeLimit: asn1::ASN1Message(True, self.depth) &convert=$$.body.num_value &default=0;
+  timeLimit: asn1::ASN1Message(True, self.depth) &convert=$$.body.num_value &default=0;
+  typesOnly: asn1::ASN1Message(True, self.depth) &convert=$$.body.bool_value &default=False;
+  filter: SearchFilter(self.depth);
+  attributes: AttributeSelection(self.depth);
+
+  on %done {
+    message.obj = self.baseObject;
+    message.arg = "%s" % self.scope;
```
bbannier, line 416: Let's move these into `%done` hooks for `baseObject` and `scope`, respectively. Once we have done that we could probably hide the files after consuming their data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+type SearchResultEntry = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  objectName: asn1::ASN1Message(True, self.depth) &convert=$$.body.octetstring;
+  # TODO: if we want to descend down into PartialAttributeList
+  attributes: asn1::ASN1Message(True, self.depth);
+
+  on %done {
+    message.obj = self.objectName;
```
bbannier, line 429: Let's move this into a `%done` hook for `objectName` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  }
+};
+
+type SearchResultDone = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
```
bbannier, line 440: Let's move this into a `%done` hook for `result` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+    print "SearchResultDone", message.messageID, self;
```
bbannier, line 441:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  }
+};
+
+type SearchResultReference = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: SearchResultReference
+  # The SearchResultReference is of the same data type as the Referral.
```
bbannier, line 449: I see no implementation for `Referral`, so the comment is currently not that helpful. If you use the parsers interchangebly below, you could use a type alias to keep their semantics straigt, e.g.,

```.r
type Referral = SearchResultReference;
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  on %done {
+    message.result = self.result;
+    print "SearchResultDone", message.messageID, self;
+  }
+};
+
+type SearchResultReference = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: SearchResultReference
```
bbannier, line 448: Is it correct that this parser will consume no data? If not, let's instead either skip over its data, or leave out of `Message` for now.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  }
+};
+
+type ModifyResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
```
bbannier, line 479: Let's move this into a `%done` hook for `result` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+type SearchResultReference = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: SearchResultReference
+  # The SearchResultReference is of the same data type as the Referral.
+
+  on %done {
+    print "SearchResultReference", message.messageID, self;
```
bbannier, line 452:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+type ModifyRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  objectName: asn1::ASN1Message(True, self.depth) &convert=$$.body.octetstring;
+
+  # TODO: parse changes
+
+  on %done {
+    message.obj = self.objectName;
+    print "ModifyRequest", message.messageID, self;
```
bbannier, line 469:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+type ModifyRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  objectName: asn1::ASN1Message(True, self.depth) &convert=$$.body.octetstring;
+
+  # TODO: parse changes
+
+  on %done {
+    message.obj = self.objectName;
```
bbannier, line 468: Let's move this into a `%done` hook for `objectName` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+type ModifyResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
+    print "ModifyResponse", message.messageID, self;
```
bbannier, line 480:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+#-----------------------------------------------------------------------------
+# Add Operation
+# https://tools.ietf.org/html/rfc4511#section-4.7
+
+type AddRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: add request sequence
```
bbannier, line 491: Is it correct that this parser will consume no data? If not, let's instead either skip over its data, or leave out of `Message` for now.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+  on %done {
+    print "AddRequest", message.messageID, self;
+  }
```
bbannier, line 495:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+type AddResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
```
bbannier, line 504: Let's move this into a `%done` hook for `result` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+    print "AddResponse", message.messageID, self;
```
bbannier, line 505:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+# Delete Operation
+# https://tools.ietf.org/html/rfc4511#section-4.8
+
+type DelRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  objectName: asn1::ASN1Message(True, self.depth) &convert=$$.body.octetstring;
+
+  on %done {
+    message.obj = self.objectName;
```
bbannier, line 519: Let's move this into a `%done` hook for `objectName` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+    print "DelRequest", message.messageID, self;
```
bbannier, line 520:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  }
+};
+
+type DelResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
```
bbannier, line 530: Let's move this into a `%done` hook for `result` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+    print "DelResponse", message.messageID, self;
```
bbannier, line 531:
```suggestion
```

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+
+#- ASN.1 bit string -----------------------------------------------------------
+# https://www.obj-sys.com/asn1tutorial/node10.html
+
+type ASN1BitString = unit(len: uint64, constructed: bool) {
+  unused_bits: uint8;
+  value_bits: bytes &size=(len - 1);
+
+  # TODO - constructed form
+  # `bytes` needs << and >> support before we can implement complex bitstrings
```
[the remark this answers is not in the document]
mmguero: ok, https://github.com/zeek/spicy/issues/921 and I'll put a comment in that code

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+# https://tools.ietf.org/html/rfc4511#section-4.8
+
+type ModDNRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse mod DN request sequence
+
+  on %done {
+    print "ModDNRequest", message.messageID, self;
+  }
```
bbannier, line 546:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+#-----------------------------------------------------------------------------
+# Modify DN Operation
+# https://tools.ietf.org/html/rfc4511#section-4.8
+
+type ModDNRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse mod DN request sequence
```
bbannier, line 542: Is it correct that this parser will consume no data? If not, let's instead either skip over its data, or leave out of `Message` for now.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+type ModDNResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
+    print "ModDNResponse", message.messageID, self;
```
bbannier, line 556:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  }
+};
+
+type ModDNResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
```
bbannier, line 555: Let's move this into a `%done` hook for `result` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+# https://tools.ietf.org/html/rfc4511#section-4.10
+
+type CompareRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse compare request entry/ava sequence
+
+  on %done {
+    print "CompareRequest", message.messageID, self;
+  }
```
bbannier, line 571:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+#-----------------------------------------------------------------------------
+# Compare Operation
+# https://tools.ietf.org/html/rfc4511#section-4.10
+
+type CompareRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse compare request entry/ava sequence
```
bbannier, line 567: Is it correct that this parser will consume no data? If not, let's instead either skip over its data, or leave out of `Message` for now.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  }
+};
+
+type CompareResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  on %done {
+    message.result = self.result;
```
bbannier, line 580: Let's move this into a `%done` hook for `result` and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+    print "CompareResponse", message.messageID, self;
```
bbannier, line 581:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  }
+};
+
+#-----------------------------------------------------------------------------
+# Abandon Operation
+# https://tools.ietf.org/html/rfc4511#section-4.11
+
+type AbandonRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+  # messageID: asn1::ASN1Message(True, self.depth) &convert=$$.body.num_value;
```
bbannier, line 591: Don't we need to at least skip over the data here?

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+  on %done {
+    # TODO: can't do string -> bytes
+    # but see https://github.com/zeek/spicy/issues/911 for next release
+    # message.obj = "%d" % (self.messageID);
+    print "AbandonRequest", message.messageID, self;
```
bbannier, line 597:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+# https://tools.ietf.org/html/rfc4511#section-4.11
+
+type AbandonRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+  # messageID: asn1::ASN1Message(True, self.depth) &convert=$$.body.num_value;
+
+  on %done {
+    # TODO: can't do string -> bytes
+    # but see https://github.com/zeek/spicy/issues/911 for next release
+    # message.obj = "%d" % (self.messageID);
```
bbannier, line 596: This should be in a `%done` hook.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+# https://tools.ietf.org/html/rfc4511#section-4.12
+
+type ExtendedRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse ExtendedRequest sequence
+
+  on %done {
+    print "ExtendedRequest", message.messageID, self;
+  }
```
bbannier, line 612:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+#-----------------------------------------------------------------------------
+# Extended Operation
+# https://tools.ietf.org/html/rfc4511#section-4.12
+
+type ExtendedRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse ExtendedRequest sequence
```
bbannier, line 608: Is it correct that this parser will consume no data? If not, let's instead either skip over its data, or leave out of `Message` for now.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+  on %done {
+    print "ExtendedRequest", message.messageID, self;
+  }
+};
+
+type ExtendedResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse ExtendedRequest sequence
```
bbannier, line 618: Is it correct that this parser will consume no data? If not, let's instead either skip over its data, or leave out of `Message` for now.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+  on %done {
+    # message.result = self.result;
```
bbannier, line 621: Let's move this into a `%done` hook for `result` (after defining it) and then potentially hide the field from users after consuming its data.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+    print "ExtendedResponse", message.messageID, self;
```
bbannier, line 622:
```suggestion
```

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+};
+
+#-----------------------------------------------------------------------------
+# IntermediateResponse Message
+# https://tools.ietf.org/html/rfc4511#section-4.13
+
+type IntermediateResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  # TODO: parse IntermediateResponse sequence
```
bbannier, line 633: Is it correct that this parser will consume no data? If not, let's instead either skip over its data, or leave out of `Message` for now.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+  on %done {
+    print "IntermediateResponse", message.messageID, self;
+  }
```
bbannier, line 637:
```suggestion
```

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+      throw "Unsupported ASN1String type";
+    }
+  }
+
+};
+
+#- ASN.1 OID ------------------------------------------------------------------
+# https://www.obj-sys.com/asn1tutorial/node124.html
+
+type oidnibble = unit {
```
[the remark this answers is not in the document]
mmguero: renamed to `ASN1ObjectIdentifierNibble`

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+  submessages: ASN1Message(True, depth)[] &eod;
+} &size=msgLen;
+
+#- ASN.1 message with header and body -----------------------------------------
+# UNIVERSAL or APPLICATION/CONTEXT_SPECIFIC/PRIVATE
+# - if UNIVERSAL, body:ASN1Body is parsed
+# - else, application_data:bytes stores data array
+
+public type ASN1Message = unit(recursive: bool, depth: uint32) {
+  var depth: uint32 = depth;
```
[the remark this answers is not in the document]
mmguero: just for the debug output. i will drop it

analyzer/protocol/ldap/asn1.spicy
mmguero: Will remove.

analyzer/protocol/ldap/ldap.evt
```diff
@@ -0,0 +1,41 @@
+protocol analyzer spicy::ldap_tcp over TCP:
+  parse with ldap::Messages,
+  ports {389/tcp, 3268/tcp};
+
+# TODO: UDP hasn't been tested/implemented yet
+# TODO: test this, will it be Message or Messages?
+# protocol analyzer spicy::ldap_udp over UDP:
+#   parse with ldap::Messages,
+#   ports {389/udp, 3268/udp};
```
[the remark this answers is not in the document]
mmguero: I'm not implementing LDAP over UDP right now. I'll address the comment.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  MOD_DN_REQUEST          = 12,
+  MOD_DN_RESPONSE         = 13,
+  COMPARE_REQUEST         = 14,
+  COMPARE_RESPONSE        = 15,
+  ABANDON_REQUEST         = 16,
+  SEARCH_RESULT_REFERENCE = 19,
+  EXTENDED_REQUEST        = 23,
+  EXTENDED_RESPONSE       = 24,
+  INTERMEDIATE_RESPONSE   = 25,
+  NOT_SET                 = 255
```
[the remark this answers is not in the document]
mmguero: That's correct, but there is a bug in the `UNDEF` enum value that makes it unusable in Zeek (see https://github.com/zeek/spicy-plugin/issues/35). I'll put a note in to remove my NOT_SET values once that bug is fixed and in a stable release.

analyzer/protocol/ldap/asn1.spicy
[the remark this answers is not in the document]
bbannier: Actually meant the empty line 😅

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+                                                    &default="";
+
+  # TODO: if we want to parse referral URIs in result
+  # https://tools.ietf.org/html/rfc4511#section-4.1.10
+};
+
+#-----------------------------------------------------------------------------
+public type Messages = unit {
+  : Message[];
+};
```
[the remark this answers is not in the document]
mmguero: Yes, it's used by the .evt file. It exists in all other TCP parsers in spicy-analyzer as well.

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+                             &convert=$$.to_uint(spicy::ByteOrder::Network) {
+      self.len = self.length_parse;
+      self.tag_len = self.data.num + 1;
+    }
+  };
+};
+
+type ASN1Tag = unit {
+  var tpe: ASN1Type;
+  var len: uint8 = 1;
```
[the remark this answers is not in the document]
bbannier: We never set a value here, so I am unsure why it is needed here.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+public type Message = unit {
+  var messageID: int64;
+  var opcode: ProtocolOpcode = ProtocolOpcode::NOT_SET;
+  var applicationBytes: bytes;
+  var unsetResultDefault: Result;
+  var result: Result& = self.unsetResultDefault;
+  var obj: string = "";
+  var arg: string = "";
```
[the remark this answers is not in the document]
mmguero: Yes, I set these values in the child units and then pass them to the event for logging in Zeek.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+      }
+      if (self.seq.body.seq.submessages[1]?.application_id) {
+        self.opcode = cast<ProtocolOpcode>(cast<uint8>(self.seq.body.seq.submessages[1].application_id));
+        self.applicationBytes = self.seq.body.seq.submessages[1].application_data;
+      }
+    }
+    if ((! self?.messageID) ||
+        (! self?.opcode) ||
+        (self.opcode == ProtocolOpcode::NOT_SET)) {
+      throw "LDAP message is not ASN.1 sequence as described in RFC 4511";
```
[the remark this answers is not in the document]
mmguero: great idea, thanks.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+  name: asn1::ASN1Message(True, self.depth) &convert=$$.body.octetstring;
+  var authType: BindAuthType = BindAuthType::NOT_SET;
+  var simpleCreds: string = "";
+
+  authentication: asn1::ASN1Message(True, self.depth) {
+    if (self.authentication?.application_id) {
+      self.authType = cast<BindAuthType>(cast<uint8>(self.authentication.application_id));
+    }
+    if ((! self?.authType) || (self.authType == BindAuthType::NOT_SET)) {
+      throw "BindRequest is not ASN.1 sequence as described in RFC 4511";
```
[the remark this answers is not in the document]
mmguero: will do

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+  on %done {
+    message.obj = self.name;
+    if (|self.simpleCreds| > 0) {
+      message.arg = self.simpleCreds;
+    } else if (self.authType == BindAuthType::BIND_AUTH_SASL) {
+      message.arg = self.saslCreds.mechanism;
+    }
+    print "BindRequest", message.messageID, self.version, self.name,
+          self.authType, message.obj, message.arg;
```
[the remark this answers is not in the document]
mmguero: ok, i will remove all the prints.

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+
+type BindResponse = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  result: Result(self.depth+1);
+
+  # TODO: if we want to parse SASL credentials returned
+
+  on %done {
+    message.result = self.result;
```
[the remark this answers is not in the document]
mmguero: great idea, thank you

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+#-----------------------------------------------------------------------------
+# Unbind Operation
+# https://tools.ietf.org/html/rfc4511#section-4.3
+
+type UnbindRequest = unit(inout message: Message) {
+  var depth: uint32 = message.seq.body.seq.submessages[1].depth+1;
+
+  on %done {
+    print "UnbindRequest", message.messageID, self;
+  }
```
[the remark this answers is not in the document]
mmguero: Great, I'll remove all of the on %done statements in favor of hooks.

analyzer/protocol/ldap/asn1.spicy
```diff
@@ -0,0 +1,339 @@
...
+    ASN1Type::GeneralizedTime,
+    ASN1Type::UTCTime,
+    ASN1Type::UTF8String,
+    ASN1Type::BMPString,
+    ASN1Type::UniversalString -> asn1string: ASN1String(head.tag, head.len.len)
+                                             &convert=$$.value;
+
+    ASN1Type::SEQUENCE, ASN1Type::SET -> seq: ASN1SubMessages(head.len.len, depth+1) if (recursive);
+
+    * -> : bytes &size=head.len.len {
```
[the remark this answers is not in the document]
bbannier: Let's just explicitly pattern-match the values by expanding `*` into cases. You do not need that to give that field a name if you do not need it; a comment is just as good (and we might actually optimize stuff like that in the future).

analyzer/protocol/ldap/ldap.spicy
```diff
@@ -0,0 +1,638 @@
...
+public type Message = unit {
+  var messageID: int64;
+  var opcode: ProtocolOpcode = ProtocolOpcode::NOT_SET;
+  var applicationBytes: bytes;
+  var unsetResultDefault: Result;
+  var result: Result& = self.unsetResultDefault;
+  var obj: string = "";
+  var arg: string = "";
+
+  seq: asn1::ASN1Message(True, 0) {
```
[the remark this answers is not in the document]
mmguero: Okay, I will go through and fix as many of those as I can find.

analyzer/protocol/ldap/ldap.spicy
mmguero: I'm going to move the done to the hook, but I think in these cases I do want to have the fields at the sub-unit level as well.

analyzer/protocol/ldap/ldap.spicy
mmguero: removed the unhelpful comment, it was from the rfc

analyzer/protocol/ldap/ldap.spicy
mmguero: In this case we have already parsed the data, we're just doing a `parse-from` the parent Message. ASN.1 already descended the tree for us. For now I've commented this out with the other unimplemented ones.

mmguero:
Finished @bbannier 's review and making changes from it. I'm going to do some more testing and will reply here once I feel things are all as stable as they were pre-review. Thanks a lot

mmguero:
Made the changes suggested by @bbannier in his review, minus the few of them that I gave reasons for in the replies to his comments. I'm still getting all the same `.log` output and passing my tests, so that's a good thing ;).

analyzer/protocol/ldap/asn1.spicy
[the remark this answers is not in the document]
mmguero: Okay, I've got those cases all expanded out now in the switch statement.

analyzer/protocol/ldap/asn1.spicy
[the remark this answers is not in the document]
mmguero: You're right, removed.

bbannier (dismissed):
Thanks for spending the time to clean this up @mmguero.

Overall this looks good to me. Like I mentioned offline we could still make this easier to consume by minimizing data exposed by units (you often use members for parsing, and then set some dependent unit `var`; ideally we'd wouldn't expose the member used for parsing after that). If you do not have time for that cleanup, I believe we take that on after merging.

Let's also see whether @rsmmr has any additional suggestions.

bbannier (approved):

mmguero:
Something got a little bungled in merging from upstream, I'll fix whatever it is and comment here again when I'm good-to-go.

mmguero:
Okay, I got it sorted. I think I'm good whenever @bbannier and/or @rsmmr are.

rsmmr:
This has come together nicely, good job!

I went ahead and merge it, and created a ticket tracking some TODOs: https://github.com/zeek/spicy-analyzers/issues/58

Also, I turned the PR's description into a README. @mmguero Feel free to file another PR if you want clean that up a bit more.

PR #61, opened 2021-05-31
Implement LDAP analyzer enhancement after upstream changes

### LINKED ISSUE #58: [LDAP] Pending enhancements
A few things pending still for the LDAP analyzer, waiting for other tickets:

- [ ] Once &convert is fixed for unit types with parameters, we can apply it to more types. (https://github.com/zeek/spicy/issues/917)
- [ ]  Once we have `Undef` Zeek-side, use it to replace the `NOT_SET` fields (https://github.com/zeek/spicy-plugin/issues/35)
- [ ]  Once we have switch-level `&parse-from`, use that instead of one per case (https://github.com/zeek/spicy/issues/913)

bbannier:
Closes #58.

### DIFF
--- a/analyzer/protocol/ldap/asn1.spicy
+++ b/analyzer/protocol/ldap/asn1.spicy
@@ -129,11 +129,10 @@ type ASN1OctetString = unit(len: uint64, constructed: bool) {
 # https://www.obj-sys.com/asn1tutorial/node124.html
 
 type ASN1String = unit(tag: ASN1Tag, len: uint64) {
-  var value: string = "";
+  var encoding: hilti::Charset;
 
-  : ASN1OctetString(len, tag.constructed) {
+  on %init {
     switch ( tag.type_ ) {
-
       # see "Restricted Character String Types" in
       # "Generic String Encoding Rules (GSER) for ASN.1 Types"
       # (https://datatracker.ietf.org/doc/html/rfc3641#section-3.2)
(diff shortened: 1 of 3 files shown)


analyzer/protocol/ldap/asn1.spicy
```diff
@@ -129,19 +129,18 @@ type ASN1OctetString = unit(len: uint64, constructed: bool) {
...
     switch ( tag.type_ ) {
-
       # see "Restricted Character String Types" in
       # "Generic String Encoding Rules (GSER) for ASN.1 Types"
       # (https://datatracker.ietf.org/doc/html/rfc3641#section-3.2)
 
       case ASN1Type::PrintableString,
            ASN1Type::GeneralizedTime,
            ASN1Type::UTCTime: {
-        self.value = $$.value.decode(hilti::Charset::ASCII);
```
rsmmr, line 144: Might be more convenient to keep the decoding, so that the user of the type doesn't need to worry about that charset anymore.

rsmmr (approved):

analyzer/protocol/ldap/asn1.spicy
bbannier: We now decode this after extracting `value`, and convert the whole unit to a string with a unit convert. The user should never need to worry how `ASN1String`s are encoded.

rsmmr: Ah, I see it now, nevermind.

PR #66, opened 2021-06-16
Tailscale VPN

keithjjones:
Tailscale's VPN takes some components of Wireguard and adds other data in a manner that it won't detect using our Wireguard analyzer.  Therefore I wrote this so we could detect Tailscale's usage of Wireguard.  There is replication in the tailscale.spicy file with the original wireguard.spicy file, but I don't know of a smooth way around that.


0xxon:
This looks very interesting - I was not aware of tailscale before.

I wonder if it is possible to do this without replicating the entire wireguard analyzer again - at least to me that seems somewhat wasteful - as you already noted too. If they really just wrap wireguard in their packets - you should actually just be able to import the wireguard analyzer into the tailscale analyzer and use the wireguard units in it - I think. That should get around the copy/pasting.

keithjjones:
Ok, how do I go about doing that cross reference?

bbannier:
> Ok, how do I go about doing that cross reference?

In order to make the wireguard module visible in your tailscale analyzer add the following before `spicy_add_analyzer` in `/analyzer/protocol/tailscale/CMakeLists.txt`,

```.cmake
set(SPICYZ_FLAGS ${SPICYZ_FLAGS} -L${CMAKE_CURRENT_SOURCE_DIR}/../wireguard)
```

The you can `import Wireshark;` in `tailscale.spicy`.

The major different bit seems to be the `magic` the outer tailscale layer injects into wireguard packages. You should be able to reuse the original wireguard analyzer if you add a wireguard parser to your `TailScalePacket` as a `sink`, and `write` `TailScalePacket`'s `magic` to it after parsing `disc` (e.g., in `disc`'s field hook). You should be able to guard the `write` with the same conditional you already use on your existing `wg` field, and `close` the `sink` otherwise).

keithjjones:
> > Ok, how do I go about doing that cross reference?
> 
> In order to make the wireguard module visible in your tailscale analyzer add the following before `spicy_add_analyzer` in `/analyzer/protocol/tailscale/CMakeLists.txt`,
> 
> ```cmake
> set(SPICYZ_FLAGS ${SPICYZ_FLAGS} -L${CMAKE_CURRENT_SOURCE_DIR}/../wireguard)
> ```
> 
> Then you can `import Wireshark;` in `tailscale.spicy`.
> 
> The major different bit seems to be the `magic` the outer tailscale layer injects into wireguard packages. You should be able to reuse the original wireguard analyzer if you add a `Wireguard` parser to your `TailScalePacket` as a `sink`, and `write` `TailScalePacket`'s `magic` to it after parsing `disc` (e.g., in `disc`'s field hook). You should be able to guard the `write` with the same conditional you already use on your existing `wg` field, and `close` the `sink` otherwise).

This is far more time than I have to give to this.  If you'd like to take a stab at this, I'll merge into my merge request.  I thought it was going to be easier.

keithjjones:
Changes pushed.  Thanks!

analyzer/protocol/wireguard/wireguard.spicy
```diff
@@ -30,7 +30,7 @@ public type WireGuardPacket = unit {
 	};
 };
 
-type HandshakeInitiation = unit {
+public type HandshakeInitiation = unit {
```
bbannier, line 33: These shouldn't be `public`, I'll remove this on merge.

bbannier (approved):

Issue #65, opened 2021-06-10
STUN Spicy Analyzer


keithjjones:
This repo supports STUN:  https://github.com/r-franke/spicy_stun

I can probably get this working and added to spicy-analyzers.  I will add an issue at that repo pointing here and see if we can merge the work here.

More info:
https://datatracker.ietf.org/doc/html/rfc5389

PR #67, opened 2021-06-17
Added STUN protocol

keithjjones:
Got things working from https://github.com/r-franke/spicy_stun/issues/1 and added some logic too.

Testing this on 7 networks and it looks good.  More info in readme.md in the stun directory.

This will close https://github.com/zeek/spicy-analyzers/issues/65


analyzer/protocol/stun/Readme.md
```diff
@@ -0,0 +1,58 @@
+# STUN
+
+Original logic can be found here:
+
+- https://github.com/r-franke/spicy_stun
+- https://github.com/r-franke/spicy_stun/issues/1
+
+Additional logic has been added on top of the original logic.
```
bbannier, line 8: Could you first put a block explaining what this analyzer does (e.g., "... adds an analyzer for the STUN protocol, see ... "). You could then call out that this is based on r-frankie/spicy-stun, but I would drop the "Additional logic has been added ..." part.

While you are at it, could you rename this file `README` most other files?

analyzer/protocol/stun/consts.zeek
```diff
@@ -0,0 +1,85 @@
...
+		[0xC057] = "GOOG_NETWORK_INFO",
+		[0xC058] = "GOOG_LAST_ICE_CHECK_RECEIVED",
+		[0xC059] = "GOOG_MISC_INFO",
+		[0xC05A] = "GOOG_OBSOLETE_1",
+		[0xC05B] = "GOOG_CONNECTION_ID",
+		[0xC05C] = "GOOG_DELTA",
+		[0xC05D] = "GOOG_DELTA_ACK",
+		[0xC060] = "GOOG_MESSAGE_INTEGRITY_32",
+	} &default = function(n: count): string { return fmt("unknown-attrtype-%d", n); };
+
```
bbannier, line 84:
```suggestion
```

analyzer/protocol/stun/stun.evt
```diff
@@ -0,0 +1,101 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
```
bbannier, line 1: Can you also fix uses of tabs for alignment here as well?

analyzer/protocol/stun/stun.evt
```diff
@@ -0,0 +1,101 @@
...
+
+protocol analyzer spicy::STUN over UDP:
+	parse with STUN::STUNPacket;
+
+import STUN;
+
+on STUN::STUNPacket -> event STUN::STUNPacket($conn, $is_orig, self.M, self.C, self.trans_id);
+
+# SOFTWARE
+on STUN::Attribute::DESCRIPTION -> event STUN::string_attribute($conn, $is_orig, self.method,
```
bbannier, line 11: Can you explain why we trigger separate events for each attribute? This analyzer should only parse `STUNPacket` which already contain all the `Attribute`s.

Even if this currently doesn't e.g., produce double logging, I wonder if it wouldn't be simpler to make the handler for `STUN::STUNPacket` more powerful, so we should have a good reason for this complication.

analyzer/protocol/stun/stun.spicy
```diff
@@ -0,0 +1,215 @@
...
+	  m1: 5..7;
+	  c1: 8;
+	  m2: 9..13;
+	  zeros: 14..15 &requires=($$ == 0);
+	};
+
+	on msgtype {
+		self.M = self.msgtype.m0 + 16*self.msgtype.m1 + 128*self.msgtype.m2;
+		self.C = self.msgtype.c0 + 2*self.msgtype.c1;
+	}
```
bbannier, line 136: We don't need `msgtype` after filling in `M` and `C`, let's remove that additional state,

```suggestion
	: bitfield(16) {
	  m0: 0..3;
	  c0: 4;
	  m1: 5..7;
	  c1: 8;
	  m2: 9..13;
	  zeros: 14..15 &requires=($$ == 0);
	} {
		self.M = self.msgtype.m0 + 16*self.msgtype.m1 + 128*self.msgtype.m2;
		self.C = self.msgtype.c0 + 2*self.msgtype.c1;
	}
```

analyzer/protocol/stun/stun.spicy
```diff
@@ -0,0 +1,215 @@
...
+#      |                            ...]                               |
+#      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+#
+# 		msg_type:
+#                       0                 1
+#                       2  3  4 5 6 7 8 9 0 1 2 3 4 5
+#                      +--+--+-+-+-+-+-+-+-+-+-+-+-+-+
+#                      |M |M |M|M|M|C|M|M|M|C|M|M|M|M|
+#                      |11|10|9|8|7|1|6|5|4|0|3|2|1|0|
+#                      +--+--+-+-+-+-+-+-+-+-+-+-+-+-+
```
bbannier, line 118:
```suggestion
#  RFC5389:
#
# Magic cookie: 0x2112A442
#
#  0                   1                   2                   3
#  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
# +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
# |0 0|     STUN Message Type     |         Message Length        |
# +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
# |                         Magic Cookie                          |
# +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
# |                                                               |
# |                     Transaction ID (96 bits)                  |
# |                                                               |
# +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
# |                        [Attributes,                           |
# |                            ...]                               |
# +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
#
#  msg_type:
#                  0                 1
#                  2  3  4 5 6 7 8 9 0 1 2 3 4 5
#                 +--+--+-+-+-+-+-+-+-+-+-+-+-+-+
#                 |M |M |M|M|M|C|M|M|M|C|M|M|M|M|
#                 |11|10|9|8|7|1|6|5|4|0|3|2|1|0|
#                 +--+--+-+-+-+-+-+-+-+-+-+-+-+-+
```

analyzer/protocol/stun/stun.spicy
```diff
@@ -0,0 +1,215 @@
...
+	  zeros: 14..15 &requires=($$ == 0);
+	};
+
+	on msgtype {
+		self.M = self.msgtype.m0 + 16*self.msgtype.m1 + 128*self.msgtype.m2;
+		self.C = self.msgtype.c0 + 2*self.msgtype.c1;
+	}
+
+	msg_length: uint16;
+	cookie: uint32 &requires=($$ == 554869826);
```
bbannier, line 139: This is not needed. Let's instead before this unit declare

```
const MAGIC_COOKIE = 0x2112A442;
```

and then encode
```
	: MAGIC_COOKIE;
```

With that we can also get rid of the comment on the cookie in the unit header.

analyzer/protocol/stun/stun.spicy
```diff
@@ -0,0 +1,215 @@
...
+		AttrType::XOR_PEER_ADDRESS -> XOR_PEER_ADDRESS: MappedAddress(self.attr_length);
+		AttrType::XOR_RELAYED_ADDRESS -> XOR_RELAYED_ADDRESS: MappedAddress(self.attr_length);
+		AttrType::PRIORITY -> PRIORITY: bytes &size=4;
+		AttrType::USE_CANDIDATE -> USE_CANDIDATE: bytes &size=0;
+		AttrType::SOFTWARE -> DESCRIPTION: bytes &size=self.attr_length;
+		AttrType::ALTERNATE_SERVER -> ALTERNATE_SERVER: MappedAddress(self.attr_length);
+		AttrType::FINGERPRINT -> CRC_32: bytes &size=self.attr_length;
+		AttrType::ICE_CONTROLLED -> ICE_CONTROLLED: bytes &size=self.attr_length;
+		AttrType::ICE_CONTROLLING -> ICE_CONTROLLING: bytes &size=self.attr_length;
+		* -> attr_value: bytes &size=self.total_size;
```
bbannier, line 180: Let's add a TODO if this is for currently unhandled attributes.

analyzer/protocol/stun/stun.spicy
```diff
@@ -0,0 +1,215 @@
...
+	};
+	padding: bytes &size=self.value_padding;
+
+	var value_padding: uint16;
+	var total_size: uint16;
```
bbannier, line 185: Let's move these up to the other `var` decls so it isn't hidden down here.

tests/protocol/stun/stun.zeek
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+
+# @TEST-EXEC: set >envs
+# @TEST-EXEC: ${ZEEK} -NN >zeek 2>&1
+# @TEST-EXEC: ${ZEEK} -Cr ${TRACES}/stun-ice-testcall.pcap %INPUT
```
bbannier, line 5: We need to document the provenience of this pcap file.

tests/protocol/stun/stun.zeek
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+
+# @TEST-EXEC: set >envs
```
bbannier, line 3: Let's drop this debugging command.
```suggestion
```

tests/protocol/stun/stun.zeek
```diff
@@ -0,0 +1,33 @@
...
+# @TEST-EXEC: ${ZEEK} -NN >zeek 2>&1
```
bbannier, line 4: This line currently does nothing and looks like a debugging leftover. Let's add a proper `availability` check like we do for other analyzers.

bbannier (changes requested):

0xxon:
Unless you got special explicit dispensation, we also have to include the original license somewhere (even though it also is a 3-clause BSD license).

analyzer/protocol/stun/Readme.md
keithjjones: I don't understand why this needs to be changed.  There are links to what STUN is.

analyzer/protocol/stun/stun.evt
keithjjones: I don't understand what you mean.  Each type is different and has to be parsed differently depending on the attr type.

analyzer/protocol/stun/stun.spicy
keithjjones: I don't know what you mean.

tests/protocol/stun/stun.zeek
keithjjones: It's from the original repo, that I already documented.

tests/protocol/stun/stun.zeek
keithjjones: I copied this from @0xxon Wireguard analyzer.

analyzer/protocol/stun/stun.evt
keithjjones: There isn't a tab here.

analyzer/protocol/stun/stun.spicy
keithjjones: They are handled by the *.

tests/protocol/stun/stun.zeek
bbannier: Could you add an entry to `tests/Traces/README`?

keithjjones:
> Unless you got special explicit dispensation, we also have to include the original license somewhere (even though it also is a 3-clause BSD license).

Ok, tell me where.

tests/protocol/stun/stun.zeek
keithjjones: The STUN readme isn't enough?

keithjjones:
> > Unless you got special explicit dispensation, we also have to include the original license somewhere (even though it also is a 3-clause BSD license).
> 
> Ok, tell me where.

Plus I got explicit permission if you check the link I left in the readme:  https://github.com/r-franke/spicy_stun/issues/1

analyzer/protocol/stun/stun.evt
bbannier: I meant in this file.

analyzer/protocol/stun/stun.evt
bbannier: We raise an event when we parsing any `Attribute`, but also when parsing a `STUNPacket`. `Attribute`s are always inside a `STUNPacket`. This means that for each `STUNPacket` (the top-level parsing unit) we raise an event for the full package, and then again for each of their `attributes`.

I was wondering whether we couldn't instead raise an event only for `STUNPacket` and do any Zeek-side, attribute-specific processing starting from their `attributes`.

analyzer/protocol/stun/stun.evt
keithjjones: I did.  They are all tabs.

analyzer/protocol/stun/stun.spicy
bbannier: Let's please document if there is more work to do.

analyzer/protocol/stun/stun.spicy
keithjjones: moved

analyzer/protocol/stun/Readme.md
keithjjones: I changed some of the language here.

analyzer/protocol/stun/stun.evt
keithjjones: That would probably do the same thing.  I was a) reusing parts of the original code when I wrote this and this is how it was, and b) it just made sense in my head this way.  This would be quite a bit of code rewriting, if I'm understanding you correctly.

analyzer/protocol/stun/stun.spicy
keithjjones: I wasn't planning on doing more work until I needed something.  All of the data is being sent to the log, the only change would be if someone parsed that data differently before it goes to the log?

tests/protocol/stun/stun.zeek
keithjjones: Added

analyzer/protocol/stun/Readme.md
keithjjones: Also, I like the .md because you have formatting for logs and things.

keithjjones: "Additional logic has been added" part is to signify we used code, but it's not a straight up copy and paste to here.  On another analyzer in this repo Robin had me specify something similar to this where I copied portions and wrote portions, so that was my reasoning for that language.

analyzer/protocol/stun/stun.spicy
keithjjones: This actually broke the code.  I had to add msgtype: back in before the colon.

0xxon:
> > > Unless you got special explicit dispensation, we also have to include the original license somewhere (even though it also is a 3-clause BSD license).
> > 
> > 
> > Ok, tell me where.

That is a wonderful question. I would perhaps just paste the original license on top of the source-file. 

> Plus I got explicit permission if you check the link I left in the readme: [r-franke/spicy_stun#1](https://github.com/r-franke/spicy_stun/issues/1)

I saw that. This only seems to mention the fact that you can include the source-code here. It does not seem to explicitly mention that we don't have to abide by the original license, though. The original license says that we have to include the copyright notice here too. Since the license is the same as ours, we might be able to squeeze by by just including the original copyright - and then referring to the top-level 3-clause BSD license.

Alternatively - if he gives us permission to include it without his copyright mark and just a backreference to his repo, that would also be ok.

tests/protocol/stun/stun.zeek
0xxon: That is true :).

At least when I wrote wireguard - I also just stole it from another test because it made debugging easier.

analyzer/protocol/stun/Readme.md
keithjjones: Outdated

analyzer/protocol/stun/stun.spicy
keithjjones: Added

keithjjones:
> > > > Unless you got special explicit dispensation, we also have to include the original license somewhere (even though it also is a 3-clause BSD license).
> > > 
> > > 
> > > Ok, tell me where.
> 
> That is a wonderful question. I would perhaps just paste the original license on top of the source-file.
> 
> > Plus I got explicit permission if you check the link I left in the readme: [r-franke/spicy_stun#1](https://github.com/r-franke/spicy_stun/issues/1)
> 
> I saw that. This only seems to mention the fact that you can include the source-code here. It does not seem to explicitly mention that we don't have to abide by the original license, though. The original license says that we have to include the copyright notice here too. Since the license is the same as ours, we might be able to squeeze by by just including the original copyright - and then referring to the top-level 3-clause BSD license.
> 
> Alternatively - if he gives us permission to include it without his copyright mark and just a backreference to his repo, that would also be ok.

I added to a file called "Original_License.txt".

analyzer/protocol/stun/stun.spicy
keithjjones: Added

r-franke:
> > > > Unless you got special explicit dispensation, we also have to include the original license somewhere (even though it also is a 3-clause BSD license).
> > > 
> > > 
> > > Ok, tell me where.
> 
> That is a wonderful question. I would perhaps just paste the original license on top of the source-file.
> 
> > Plus I got explicit permission if you check the link I left in the readme: [r-franke/spicy_stun#1](https://github.com/r-franke/spicy_stun/issues/1)
> 
> I saw that. This only seems to mention the fact that you can include the source-code here. It does not seem to explicitly mention that we don't have to abide by the original license, though. The original license says that we have to include the copyright notice here too. Since the license is the same as ours, we might be able to squeeze by by just including the original copyright - and then referring to the top-level 3-clause BSD license.
> 
> Alternatively - if he gives us permission to include it without his copyright mark and just a backreference to his repo, that would also be ok.

Sorry for the late response, I somehow missed the tags. Feel free to include it without the copyright mark. Excellent work @keithjjones.

PR #70, opened 2021-06-30
Update for plugin reorganisation.

rsmmr:
Also aligning scripts that are shared with the plugin so that both are
using identical versions. This is in preparation for the analyzers
package to later switch to using the versions provided by the plugin
directly.

Also fixing a version issue where scripts for the ZIP analyzer could end
up being loaded even when we didn't compile the analyzer.


ci/Dockerfile
```diff
@@ -10,14 +10,15 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
-ARG ZEEK_VERSION=3.0.11-0
+ARG ZEEK_VERSION=4.0.2-0
 
 CMD ["sh"]
 ENV DEBIAN_FRONTEND=noninteractive
 ENV PATH="/opt/spicy/bin:/opt/zeek/bin:${PATH}"
 
 RUN apt-get update \
  && apt-get install -y --no-install-recommends curl ca-certificates gnupg2 less sudo \
+ && pip3 install --no-cache-dir pre-commit \
```
bbannier, line 21: If we want to execute `pip3` here we need to move the install of `python3-pip` (and maybe `python` and `python3` as well) from l.26 to the `apt-get install` invocation above.

rsmmr: Fixed

.cirrus.yml
```diff
@@ -12,7 +19,7 @@ package_ci_task:
         - ZEEK_VERSION: 3.0.13-0
       - docker_arguments:
         - ZEEK_LTS: 1
-        - ZEEK_VERSION: 4.0.0-0
+        - ZEEK_VERSION: 4.0.2-0
```
bbannier, line 22: Might just go with 4.0.3 now.

.cirrus.yml
```diff
@@ -1,8 +1,15 @@
 environment:
     # Enforce sequential JIT'ing of files for controlled memory usage.
     HILTI_JIT_SEQUENTIAL: 1
+    SPICY_PLUGIN_BRANCH: topic/robin/plugin-reorg
```
bbannier, line 4: Let's fix this before merging.

.cirrus.yml
```diff
@@ -82,14 +89,14 @@ macos_ci_task:
...
-standalone_ci_task:
+standalone_ubuntu_task:
   timeout_in: 120m
   container:
     dockerfile: ci/Dockerfile
     matrix:
       - docker_arguments:
         - ZEEK_LTS:
-        - ZEEK_VERSION: 4.0.0-0
+        - ZEEK_VERSION: 4.0.2-0
```
bbannier, line 99: 4.0.3

analyzer/CMakeLists.txt
```diff
@@ -3,15 +3,8 @@
...
 
-if ( SPICY_SCRIPTS_OUTPUT_DIR_INSTALL )
-    if ( SPICY_IN_TREE_BUILD )
-        # Do not install by default but tie to install-spicy-analyzers target.
-        install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/ DESTINATION "${SPICY_SCRIPTS_OUTPUT_DIR_INSTALL}/spicy-analyzers/"
-            COMPONENT spicy-analyzers EXCLUDE_FROM_ALL
-            FILES_MATCHING PATTERN "*.zeek" PATTERN "*.sig")
-    else ()
-        install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/ DESTINATION "${SPICY_SCRIPTS_OUTPUT_DIR_INSTALL}/spicy-analyzers"
+# TODO: Should probably switch to listing the scripts to be installed.
```
bbannier, line 6: Let's track this in a ticket or fix now.

analyzer/protocol/ldap/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 # Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
 
-spicy_add_analyzer(LDAP ldap.spicy ldap_zeek.spicy ldap.evt)
+# FIXME: This currently does not pass tests.
```
bbannier, line 3: Let's fix this before merging.

analyzer/protocol/ldap/__load__.zeek
```diff
@@ -1,3 +1,4 @@
 # Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
 
-@load ./main
+# Temporarily disabled.
```
bbannier, line 3: Let's fix this before merging.

ci/Dockerfile
```diff
@@ -10,19 +10,20 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 ARG SKIP_BUILD=
 ARG ZEEK_LTS=1
-ARG ZEEK_VERSION=3.0.11-0
+ARG ZEEK_VERSION=4.0.2-0
```
bbannier, line 13: 4.0.3

ci/Dockerfile
bbannier: Are all these actually needed here? It looks like you moved also e.g., btest dependencies here. While this works this makes it quiet messy and hard to know which step introduced which dependency (and e.g., if we'd now drop the Zeek installation below we'd likely forget to update the list here).

But probably not something which _must_ be fixed now.

PR #71, opened 2021-07-02
Added OpenVPN HMAC MD5, SHA256 and SHA512 variants 

keithjjones:
There are other HMAC sizes that I was missing.  Now this will catch connections with MD5, SHA1, SHA256, and SHA512 HMAC sizes.  We only supported the default (SHA1) with the original code.  I made the pcap for the SHA256 test.


tests/Baseline/protocol.openvpn.openvpnhmac/conn.log
```diff
@@ -9,6 +9,6 @@
 #fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	local_orig	local_resp	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents
 #types	time	string	addr	port	addr	port	enum	string	interval	count	count	string	bool	bool	count	string	count	count	count	count	set[string]
 #close XXXX-XX-XX-XX-XX-XX
-XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.56.103	33198	192.168.56.102	1194	udp	spicy_openvpn_udp_hmac	61.136881	10040	11479	SF	-	-	0	Dd	111	13148	109	14531	-
-XXXXXXXXXX.XXXXXX	ClEkJM2Vm5giqnMf4h	192.168.56.104	35701	192.168.56.102	1194	udp	spicy_openvpn_udp_hmac	59.861846	9985	11479	SF	-	-	0	Dd	110	13065	109	14531	-
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.56.103	33198	192.168.56.102	1194	udp	spicy_openvpn_udp_hmac_sha1	61.136881	10040	11479	SF	-	-	0	Dd	111	13148	109	14531	-
```
bbannier, line 12: I think long term introducing new parsers for each new combination auth scheme is not good as changes to this long are from a user perspective API breaks. Ideally we'd have a single analyzer for OpenVPN.

I filed #72 to track that cleanup.

bbannier (approved):
LGTM. We seem to be reaching a point here where we should clean up the analyzers generated for OpenVPN; I filed #72 for that.

Issue #46, opened 2021-04-07
Look into DNS false positives for OpenVPN detector


keithjjones:
I'm seeing a few false positives with dns on a busy university network and a commercial company network:

```
$ cat conn.log | python ~/Source/zeek-sanitize/sanitize.py 2,4
1617725133.905938	Cporeo3QsXCzgKXKY1	REDACTED-1	53847	REDACTED-2	53	udp	dns,spicy_openvpn_udp_hmac	-	-	-	S0	-	-	0	D	1	71	0	0	-	F	-	-	-	-	US	US	-	4020	-
1617725239.404398	CocZRy4L3lrODcxRu2	REDACTED-3	35761	REDACTED-2	53	udp	dns,spicy_openvpn_udp_hmac	-	-	-	S0	-	-	0	D	1	72	0	0	-	F	-	-	-	-	US	US	-	4020	-
1617725292.306323	CHMIrI1rocG1RPK2E7	REDACTED-4	60848	REDACTED-5	53	udp	dns,spicy_openvpn_udp_hmac	-	-	-	S0	-	-	0	D	1	71	0	0	-	F	-	-	-	-	-	US	-	4020	-
1617725372.270816	Cj4ncuA8sCnDqHd69	REDACTED-3	43049	REDACTED-2	53	udp	dns,spicy_openvpn_udp_hmac	-	-	-	S0	-	-	0	D	1	72	0	0	-	F	-	-	-	-	US	US	-	4020	-
1617725546.742503	Cqv7YX1R6ym7VLjLHj	REDACTED-6	53871	REDACTED-7	53	udp	dns,spicy_openvpn_udp_hmac	-	-	-	SHR	-	-	0	^d	0	0	1	70	-	F	-	-	-	-	-	US	-	4020	-
1617725546.742437	Cc4w7v1An2rJ4Kc2q3	REDACTED-8	50675	REDACTED-7	53	udp	dns,spicy_openvpn_udp_hmac	-	-	-	SHR	-	-	0	^d	0	0	1	70	-	F	-	-	-	-	US	US	-	4020	-
```

[...]

PR #73, opened 2021-07-06
Fixing false positives discussed in issue #46.

keithjjones:
I made the signatures require a return packet and it made these false positives stop on the networks I monitor.  

Fixes:  https://github.com/zeek/spicy-analyzers/issues/46


keithjjones:
The macos_ci error is due to a Homebrew error:

image

keithjjones:
Tests pass on my macbook:

```
$ sudo zkg test .
Password:
/Users/[REDACTED]/Source/spicy-analyzers: all tests passed```

rsmmr:
Yeah, we have that problem with Homebrew across the board currently unfortunately.

bbannier (approved):

PR #75, opened 2021-07-09
Fix trans_id indexing with STUN analyzer

keithjjones:
I messed up the indexing on the IPV6 logic and it would throw an error in reporter.log.  This MR will fix it.

### DIFF
--- a/analyzer/protocol/stun/main.zeek
+++ b/analyzer/protocol/stun/main.zeek
@@ -143,9 +143,9 @@ event STUN::mapped_address_attribute(c: connection, is_orig: bool, method: count
 			{
 			local cts = addr_to_counts(x_addr);
 			cts[0] = cts[0] ^ 0x2112A442;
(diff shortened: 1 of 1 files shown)


keithjjones:
The macos_ci test is failing because of homebrew.  The tests pass on my macbook.

The other tests fail because of a wireguard testing error.  I didn't touch anything wireguard, so those tests failing on this commit doesn't make sense.  Everything passes on my macbook:

```
$ sudo zkg test .
Password:
/Users/[REDACTED]/Source/spicy-analyzers: all tests passed
```

rsmmr:
> The macos_ci test is failing because of homebrew. The tests pass on my MacBook.

yeah, this is breaking across the board at the moment, there's a Cirrus problem with the base image unfortunately.

> The other tests fail because of a wireguard testing error.

This is likewise something that has started to appear elsewhere, needs investigation.

I'll merge.

PR #76, opened 2021-07-14
Fix for LDAP

rsmmr:


rsmmr:
This should be fixed by #76.

PR #83, opened 2021-09-09
Explicitly list scripts to install.

### LINKED ISSUE #77: Specify Zeek scripts to install explicitly.
Right now we do a blanket install of all `*.{zeek,sig}`. From `analyzer/CMakeLists.txt`:

```
# TODO: Should probably switch to listing the scripts to be installed.
install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
            DESTINATION "${SPICY_SCRIPTS_OUTPUT_DIR_INSTALL}/spicy-analyzers"
            FILES_MATCHING PATTERN "*.zeek" PATTERN "*.sig"
        )
```

bbannier:
Closes #77.


cmake/SpicyAnalyzer.cmake
```diff
@@ -0,0 +1,26 @@
+function(spicy_analyzer)
+    set(options)
+    set(oneValueArgs NAME KIND)
```
rsmmr, line 3: I don't think we want people to have to specify a "kind". Seems the need for this is just an artifact of the current structure of `spicy-analyzers`. In a standalone analyzer one wouldn't really have that.

cmake/SpicyAnalyzer.cmake
```diff
@@ -0,0 +1,26 @@
...
+
+    if (NOT DEFINED SPICY_ANALYZER_NAME)
+        message(FATAL_ERROR "NAME is required")
+    endif()
+
+    if (NOT DEFINED SPICY_ANALYZER_KIND)
+        message(FATAL_ERROR "KIND is required")
+    endif()
+
+    spicy_add_analyzer("${SPICY_ANALYZER_NAME}" "${SPICY_ANALYZER_SOURCES}")
```
rsmmr, line 16: To avoid confusion, we should probably deprecate `spicy_add_analyzer` and eventually either remove it or make it internal.

rsmmr (dismissed):
I like the new function that allows to specify all the pieces.

bbannier:
This patch needs zeek/spicy-plugin#61.

cmake/SpicyAnalyzer.cmake
bbannier: This was purely an artifact of how we organized the source tree. I got rid of the extra `file` and `protocol` level under `analyzers`, so now we can do away with `KIND` and make this function available in general in zeek/spicy-plugin; I opened zeek/spicy-plugin#61 there.

cmake/SpicyAnalyzer.cmake
bbannier: When adding this function to zeek/spicy-plugin in zeek/spicy-plugin#61 I added a mode where one can use either the new form with named args, or the previous form with unnamed args. This should help avoid any confusion.

rsmmr (approved):

PR #87, opened 2021-09-19
Add pre-commit hooks for CMake files.

bbannier:


cmake/FindSpicyPlugin.cmake
```diff
@@ -14,51 +12,49 @@
...
     endif ()
-endfunction()
+endfunction ()
 
-###
-### Main
-###
+#
+# Main
+#
```
rsmmr, line 38: I thought we had configured it to not touch comments?

analyzer/dhcp/CMakeLists.txt
```diff
@@ -1,7 +1,13 @@
...
+    NAME
+    DHCP
+    PACKAGE_NAME
+    ${PACKAGE_NAME}
+    SOURCES
+    dhcp.spicy
+    dhcp.evt
+    zeek_dhcp.spicy
+    SCRIPTS
+    __load__.zeek)
```
rsmmr, line 13: I don't like this formatting at all ... Any chance to configure this differently?

cmake/FindSpicyPlugin.cmake
```diff
@@ -59,10 +59,10 @@ if (SPICYZ)
     message(STATUS "Zeek plugin version: ${SPICY_PLUGIN_VERSION}")
 
     run_spicycz(spicy_plugin_path "--print-plugin-path")
-    set(spicy_plugin_cmake_path "${spicy_plugin_path}/cmake")
-    message(STATUS "Zeek plugin CMake path: ${spicy_plugin_cmake_path}")
+    set(SPICY_PLUGIN_CMAKE_PATH "${spicy_plugin_path}/cmake")
```
rsmmr, line 63: another thing we had decided not too touch.

PR #89, opened 2021-09-30
Copy over Zeek DPS signatures for analyzers we replace.

### LINKED ISSUE #88: HTTP analyzer does not produce any log for traffic on non-standard port
@mmguero reported the following issue [on Slack](https://zeekorg.slack.com/archives/C012CQV1K29/p1632951374043800):

>  I generated a sample PCAP and put together a simple test and noticed that the http.log generated by Zeek for that PCAP is only generated when using Zeek's built-in HTTP analyzer (either through Spicy::disable_protocol_analyzer(Analyzer::ANALYZER_SPICY_HTTP); or by doing zkg remove spicy/spicy-analyzers). When using the zeek/spicy-analyzers HTTP analyzer, the HTTP traffic doesn't get detected at all.

I looked into this and was able to reproduce the issue with a zeek-4.1.1 from Homebrew, and the most recent releases of spicy, spicy-plugin and spicy-analyzers. When I rebuild a zeek-4.1.1 from source (in `DEBUG` mode for `-B dpd` support), the issue went away.

We should look into this. It might be that e.g., Zeek-builtin DPDP for HTTP doesn't trigger for our HTTP analyzer which claims to `replaces HTTP`.

bbannier:
When specifying `replaces` in the analyzer glue we currently deactivate
the replaced analyzer before activating the replacement. We do however
not update the names referenced in DPD signatures, see e.g.,
zeek/spicy-plugin#69.

In order to activate our replacement analyzers in the same scenarios
builtin analyzers would have been activated by DPD, this patch copies
over the zeek-4.1.1 signatures of all replacement analyzers where the
replaced analyzers could have been activated with DPD in Zeek (HTTP and
DHCP).

Closes #88.


analyzer/dhcp/dpd.sig
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+
+# Signatures are copied from Zeek.
+
+signature spicy_dhcp_cookie {
+  ip-proto == udp
+  payload /^.{236}\x63\x82\x53\x63/
+  enable "dhcp"
```
rsmmr, line 8: should that be `spicy_DHCP`?

rsmmr (commented):
This works as a hot fix at least. I could have sworn that Zeek has a way to *really* replace an analyzer fully, so that `enable "xxx"` also redirects to the new analyzer. However, skimming over the code now, I'm not seeing it. So let's go with this for now.

One request: let's a add a test case at least for HTTP that makes sure DPD kicks in and activates the analyzer on non-standard ports.

rsmmr (approved):

PR #91, opened 2021-10-14
Prettier representation of STUN ERROR-CODE attribute values

mmguero:
Prettier representation of STUN ERROR-CODE attribute values

Rather than representing ERROR_CODE attr_val values like
`\x00\x00\x04-1-UNAUTHORIZED`, represent them like
`401 UNAUTHORIZED`. The class and number are now parsed
and calculated as per the RFC at
https://datatracker.ietf.org/doc/html/rfc8489#section-14.8


analyzer/stun/main.zeek
```diff
@@ -164,10 +164,17 @@ event STUN::mapped_address_attribute(c: connection, is_orig: bool, method: count
...
+								 attr_type: count, err_class: count, number: count, reason: string)
 	{
 	set_session(c);
-	local attr_val = cat(err_class, "-", number, "-", reason);
+
+	# https://datatracker.ietf.org/doc/html/rfc8489#section-14.8
+	# "The Class represents the hundreds digit of the error code. [...]
+	#  The Number represents the binary encoding of the error code module 100"
+	local err_code: count;
+	err_code = (err_class * 100) + number;
```
bbannier, line 175: While we should be forgiving in what data we accept, if `number` is outside of the accepted range (RFC8489, 14.8),

> The Number represents the binary encoding of the error code
   modulo 100, and its value MUST be between 0 and 99. 

we could produce surprising `err_code` values.

It would probably be useful to at least emit a weird if we see invalid `number`s; while we are at it we could also check whether `err_class` is in its allowed range,

> The Class represent the hundreds digit of the error code.
  The value MUST be between 3 and 6.

mmguero:
Restored the trailing spaces.

PR #93, opened 2021-11-01
Improvements in LDAP logging

fox-ds:
The changes in this pull request attempt to improve the LDAP analyzer in the following ways:

- The LDAP search filter was not parsed. Code has been added that constructs a string representation of the nested search filters, and outputs it to a `filter` field in `ldap_search.log`. 
- The search attributes were not logged. This is now possible by setting `default_log_search_attributes = T`.
- LDAP may be used in combination with SASL, but no support was yet available for this. In this PR we (naively) try to tackle this problem by first trying to parse a `Message` unit and if this fails, look for a magic byte and re-try this. In this way, one can skip the SASL layer used for signing and still parse the correct `Message` unit. The way it is implemented is similar to a try/except clause, by using the `&try` and `self.backtrack()` functions. Note that this is not the perfect solution; if the magic byte is also in the SASL layer, this will still fail. A separate SASL/GSSAPI parser might be the solution for this, in combination with keeping track of the connections that started with a 'bind SASL' operation.
- UDP is now also enabled for the analyzer, and this means that CLDAP is now also parsed. Some of the attribute values used in those search filters have different encodings/representations than the default UTF-8. Separate functions to detect some of the most common attributes are implemented.

To summarize, this PR adds the `filter` and `attributes` fields to the `ldap_search.log` and also tries to re-parse an LDAP Message unit whenever it fails the first time (e.g. because of a SASL signing)


analyzer/ldap/ldap.spicy
```diff
@@ -127,9 +127,58 @@ public type Result = unit {
...
+  }
+
+  # Again, try to parse a Message unit. Be aware that in theory (and probably in practice) this will
+  # sometimes fail if the '\x30' byte is also present in the SASL header.
+
+  # Also, we could try to do this recursively or try a few iterations, but for now I would suggest
+  # to try this extra parsing once to get the best cost/benefit tradeoff.
+  : Message &parse-from=self.remainder if ( self.success == False );
+
+};
```
bbannier, line 179: We could convert this to a `Message` when we are done so we do not expose the wrapping implementation detail outside of its concrete uses,
```suggestion
  message : Message &parse-from=self.remainder if ( self.success == False );
} &convert=self.message;
```

analyzer/ldap/main.zeek
```diff
@@ -9,6 +9,9 @@ export {
 	## Whether clear text passwords are captured or not.
 	option default_capture_password = F;
 
+  ## Whether to log LDAP search attributes or not.
+  option default_log_search_attributes = F;
```
bbannier, line 13: Let's add a test with this enabled.

To me it looks like this is currently broken. If I toggle this to `T` I see a very long `attributes` value for the test `protocol/ldap/basic.zeek` (in fact with the PCAP used in that test `ldap_search.log` now becomes >800MB big which seems wrong). I also see Zeek processes using excessive amount of memory. I suspect this behavior is due to search string building going haywire.

analyzer/ldap/ldap.spicy
```diff
@@ -127,9 +127,58 @@ public type Result = unit {
 
 #-----------------------------------------------------------------------------
 public type Messages = unit {
-  : Message[];
+  : MessageWrapper[];
 };
 
+#-----------------------------------------------------------------------------
+public type MessageWrapper = unit {
```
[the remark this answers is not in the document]
fox-ds: Yes, the 0x30 was initially found by comparing PCAPS and finding a matching byte. Moreover, WireShark's source code also mentions this special byte to identify the start of an LDAP message.

https://github.com/wireshark/wireshark/blob/eb5f4eea99593b92298bacecc5c9d885cc13a9ad/epan/dissectors/asn1/ldap/packet-ldap-template.c#L1137

Scrolling through their source code also reveals that it's not that trivial to check whether a SASL layer is present, only by following the session (check if a 'bind SASL' was performed) or even checking the first byte of the stream (if it's 0x00 they assume SASL is present).

One way to solve this is to indeed check if SASL is used and fully consume the SASL layer with a GSSAPI parser, which is currently only available as a native Zeek parser (I doubt if the native Zeek GSSAPI parser would even be able to properly parse this).

analyzer/ldap/ldap.spicy
```diff
@@ -127,9 +127,58 @@ public type Result = unit {
...
+  # strips the SASL layer if SASL Signing was enabled. Until now, I haven't found A
+  # better way to scan / determine the exact SASL header length yet, so we'll stick with this
+  # for the time being. If the entire LDAP packet was encrypted with SASL, then this will still fail and raise
+  # an error, which is expected (cause we can't decrypt encrypted SASL).
+
+  : bytes &until=b"\x30" if ( self.success == False);
+  var remainder: bytes;
+
+  # For some reason, the &until already consumes the '\x30', while it shouldn't do that.
+  # I think that's a bug in Spicy. For now, re-add it to the remainder.
```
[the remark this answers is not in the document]
fox-ds: Right, thanks for the clarification. I must have misinterpreted the docs. I'll remove the comment from the file.

analyzer/ldap/main.zeek
fox-ds: Unfortunately I can't reproduce this on my local machine, the result of my ldap_search.log is similar to the one in the CI outputs (listing only the `sAMAccountName` attribute. What command did you use on your local machine?

analyzer/ldap/main.zeek
```diff
@@ -84,6 +87,11 @@ export {
     # result diagnostic message(s)
     diagnostic_message: vector of string &log &optional;
 
+    # for now - a string representation of the 'filter' blob in the search request
+    filter: string &log &optional;
+
+    # the attributes requested in the search
+    attributes: vector of string &log &optional;
```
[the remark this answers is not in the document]
fox-ds: Sure, I'll add it.

analyzer/ldap/ldap.spicy
```diff
@@ -195,6 +254,10 @@ type SaslCredentials = unit() {
   # TODO: if we want to parse the (optional) credentials string
 };
 
+type NotImplemented = unit(inout message: Message) {
```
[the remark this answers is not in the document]
fox-ds: Yes, will add the two commented lines

analyzer/ldap/ldap.spicy
fox-ds: Yes, I changed this

fox-ds:
Will fix all the linting & pre-commit checks when all comments are resolved

analyzer/ldap/main.zeek
fox-ds: I added a test case for this that sets the attributes flag to `T` and checks if the output is a single `sAMAccountName`

bbannier: Thanks for the test.

I reinstalled my toolchain and wasn't able to reproduce the memory issue, so I suspect it was due to a faulty installation on my side.

tests/protocol/ldap/attributes.zeek
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+
+# @TEST-EXEC: zeek -C -r ${TRACES}/ldap-simpleauth.pcap %INPUT
+# @TEST-EXEC: btest-diff conn.log
+# @TEST-EXEC: btest-diff ldap.log
+# @TEST-EXEC: btest-diff ldap_search.log
+#
+# @TEST-DOC: Test LDAP analyzer with small trace.
```
bbannier, line 8:
```suggestion
# @TEST-DOC: Test LDAP search attributes with small trace.
```

analyzer/ldap/README.md
```diff
@@ -42,6 +42,8 @@ Here's what it has:
       - `base_object` (vector of 0..n search base objects specified)
       - `result_count` (number of result entries returned)
       - `result` (set of 1..n results from this uid+message_id)
+      - `filter` (search filter string)
+      - `attributes` (vector of 0..n "attributes", the attributes that were returned)
       - `diagnostic_message` (vector of 0..n diagnostic message strings)
```
bbannier, line 47: This is now neither in the order these fields appear now alphabetic order. Let's sort by the order they appear.
```suggestion
      - `diagnostic_message` (vector of 0..n diagnostic message strings)
      - `filter` (search filter string)
      - `attributes` (vector of 0..n "attributes", the attributes that were returned)
```

analyzer/ldap/ldap.spicy
```diff
@@ -305,34 +380,160 @@ type AttributeValueAssertion = unit {
...
+
+# Helper functions to properly format some custom data structures
+
+function utf16_guid_to_hex_repr(bts: bytes) : string {
+  # Rather ugly workaround to pretty-print the CLDAP DomainGuid UTF16-LE encoded string
+  # in the same format as Wireshark (aabbccdd-eeff-gghh-iijj-kkllmmnnoopp)
+
+  # We need to have exactly 16 bytes...
+  if ( |bts| != 16 ) {
+
```
bbannier, line 400:
```suggestion
```

analyzer/ldap/ldap.spicy
```diff
@@ -305,34 +380,160 @@ type AttributeValueAssertion = unit {
...
+  # Mixed little and big endian, so turn everything to big endian first...
+  # Byte 1 seems to be skipped when parsing the SID
+  for ( i in [[0], [2, 3, 4, 5, 6, 7], [11, 10, 9, 8], [15, 14, 13, 12], [19, 18, 17, 16], [23, 22, 21, 20]] ) {
+    local dec_val_rep: bytes = b"";
+    for ( j in i ) {
+      local bt: uint8 = *bts.at(j);
+      dec_val_rep += bt;
+      cnt += 1;
+      }
```
bbannier, line 454:
```suggestion
    }
```

analyzer/ldap/ldap.spicy
```diff
@@ -305,34 +380,160 @@ type AttributeValueAssertion = unit {
...
+  assertionValue: bytes &size=self.assertionValue_len;
+
+  # Only for the FILTER_EXT type, parse extra fields
+  : uint8 if ( fType == FilterType::FILTER_EXT );
+  matchValue_len: uint8 if( fType == FilterType::FILTER_EXT );
+  matchValue: bytes &size=self.matchValue_len if ( fType == FilterType::FILTER_EXT );
+
+  on %done {
+    switch ( self.attributeDesc ) {
+
```
bbannier, line 501:
```suggestion
```

analyzer/ldap/ldap.spicy
```diff
@@ -305,34 +380,160 @@ type AttributeValueAssertion = unit {
...
 };
 
 type SearchFilter = unit {
   var filterType: FilterType = FilterType::Undef;
   var filterBytes: bytes = b"";
   var filterLen: uint64 = 0;
+  var stringRepresentation: string = "";
 
   : ASN1::ASN1Message(True) {
+
```
bbannier, line 536:
```suggestion
```

analyzer/ldap/ldap.spicy
```diff
@@ -342,38 +543,139 @@ type SearchFilter = unit {
...
+    FilterType::FILTER_PRESENT -> FILTER_PRESENT: ASN1::ASN1OctetString(self.filterLen, False)
+                                                  &convert=$$.value.decode(hilti::Charset::ASCII)
                                                   &parse-from=self.filterBytes;
   };
+
+  # So when you're done with recursively parsing the filters, we can now leverage the tree structure to
+  # recursively get the stringRepresentations for those leafs, which are SearchFilters
+
+  on %done {
+
```
bbannier, line 580:
```suggestion
```

analyzer/ldap/ldap.spicy
```diff
@@ -342,38 +543,139 @@ type SearchFilter = unit {
...
+    switch ( local fType = self.filterType ) {
+
```
bbannier, line 582:
```suggestion
```

analyzer/ldap/ldap.spicy
```diff
@@ -305,34 +368,114 @@ type AttributeValueAssertion = unit {
...
+function uint32_to_hex_repr(bts: bytes) : string {
+  # Workaround to print the hex value of an uint32, prepended with '0x'
+  local ret = "0x";
+  for ( i in [3, 2, 1, 0] ) {
+    local bt: uint8 = *bts.at(i);
+    ret = ret + "%02x" % bt;
+  }
+  return ret;
+}
```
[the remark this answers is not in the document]
bbannier: What I meant here was something like this in a file like `tests/protocol/ldap/functions.spicy` (either drop the `FIXME` comments or address them),

```.r
# @TEST-EXEC: spicyc -j -d -L ${DIST}/analyzer/ldap %INPUT
#
# @TEST-DOC: Validates helper functions in LDAP module.

module test;

import LDAP;

# `utf16_guid_to_hex_repr` require exactly 16 bytes.
assert LDAP::utf16_guid_to_hex_repr(b"") == "GUID_FORMAT_FAILED";
assert LDAP::utf16_guid_to_hex_repr(b"1234567890ABCDE") == "GUID_FORMAT_FAILED";
assert LDAP::utf16_guid_to_hex_repr(b"1234567890ABCDEFG") == "GUID_FORMAT_FAILED";

# `utf16_guid_to_hex_repr` can successfully decode UTF16.
assert LDAP::utf16_guid_to_hex_repr(b"1234567890ABCDEF") == "34333231-3635-3837-3930-414243444546"; # FIXME: replace with some known-good values.

assert LDAP::bytes_sid_to_hex_repr(b"") == "";
assert LDAP::bytes_sid_to_hex_repr(b"1234567890") == "31:32:33:34:35:36:37:38:39:30"; # FIXME: replace with some known-good values.

# `bytes_sid_to_SID_repr` require exactly 24 bytes.
assert LDAP::bytes_sid_to_SID_repr(b"") == "SID_FORMAT_FAILED";
assert LDAP::bytes_sid_to_SID_repr(b"1234567890") == "SID_FORMAT_FAILED";
assert LDAP::bytes_sid_to_SID_repr(b"1234567890123456789012345") == "SID_FORMAT_FAILED";

# `bytes_sid_to_SID_repr` can successfully encode a SID representation.
assert LDAP::bytes_sid_to_SID_repr(b"123456789012345678901234") == "S-49-56299324061496-842084409-909456435-809056311-875770417"; # FIXME: replace with some known-good values.
```

bbannier (commented):
Thanks for the updates, this looks great.

I put a sample tests for the functions you added, and left trivial formatting nits, but other than that this looks ready to go for me.

analyzer/ldap/ldap.spicy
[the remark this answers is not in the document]
fox-ds: I added a bunch of tests for all 4 formatting functions

tests/protocol/ldap/attributes.zeek
fox-ds: Changed

analyzer/ldap/ldap.spicy
fox-ds: Changed

analyzer/ldap/ldap.spicy
fox-ds: Changed

analyzer/ldap/ldap.spicy
fox-ds: Changed

analyzer/ldap/ldap.spicy
fox-ds: Changed

bbannier (approved):

PR #98, opened 2021-11-08
Move LDAP analyzer to separate package.

bbannier:
This moves the LDAP analyzer into the package zeek/spicy-ldap.


.cirrus.yml
```diff
@@ -79,24 +79,24 @@ zkg_macos_task:
...
-    - pip3 install zkg btest
+    - pip3 install 'zkg>=2.12.0' btest
     - zkg autoconfig && echo "@load packages" >>"$(zeek-config --site_dir)"/local.zeek
     - git clone --branch=$ZEEK_SPICY_BRANCH --recurse-submodules https://github.com/zeek/spicy-plugin
     - zkg install --skiptests --force spicy-plugin
     - rm -rf spicy-plugin
     - zeek -N _Zeek::Spicy
 
   test_script:
-    - zkg test .
```
rsmmr, line 90: Not a blocker, but an alternative to repeating the env setup in each line could be just hardcoding things in the environment section. There shouldn't be much changing in the output of `zkg env`,

rsmmr (dismissed):
Nice that we can use the zeek-dev image now!

.cirrus.yml
bbannier: TBH, I was wondering about this as well, but I believe it would not work as we wouldn't set an env var to a static string, but compute a value from the output of a command. I'll leave this as is for now.

PR #100, opened 2021-11-17
stun: Fix integer overflow/underflow issues

awelzel:
The attribute length integer overflow was seen in the wild. The truncated
error attribute length was constructed artificially after light inspection
of the grammar.

Test pcaps are based on the stun-ice-testcall.pcap and modified with
scapy to force the attribute length to 0xffff or truncate the length
down to 0x0003 for the error attribute.

Main motivation for these fixes: The integer overflow is logged as
AnalyzerError() and makes the Zeek logs noisy due to these errors
being printed to stderr:

    analyzer/stun/stun.spicy, line 165: integer overflow


tests/protocol/stun/attribute-length-ffff.zeek
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+
```
bbannier, line 2: The format of the weird log has changed between zeek3 and zeek4. I think it would be fine to only enable this test for zeek4, e.g.,

```suggestion

# @TEST_REQUIRES: zeek-version 40000
```

This uses `zeek-version` [from spicy-plugin](https://github.com/zeek/spicy-plugin/blob/main/tests/Scripts/zeek-version).

tests/protocol/stun/malformed-error.zeek
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+
```
bbannier, line 2: Same schema issue with weird here.

analyzer/stun/stun.spicy
```diff
@@ -201,7 +201,7 @@ type MappedAddress = unit(length: uint16) {
 
 # https://tools.ietf.org/html/rfc5389#section-15.6
 type ErrorCode = unit(length: uint16) {
-	: bytes &size=2;
+	: bytes &size=2 &requires=(length >= 4);  # Pre-condition for the whole ErrorCode
```
bbannier, line 204: I think this is the only way we currently have to write this. I filed #1051 to improve this.

analyzer/stun/stun.spicy
```diff
@@ -159,11 +159,11 @@ type Attribute = unit(method: uint16, class: uint16, trans_id: bytes) {
 	var method: uint16 = method;
 	var class: uint16 = class;
 	var value_padding: uint16;
-	var total_size: uint16;
+	var total_size: uint32;
 	attr_type: uint16 &convert=AttrType($$);
 	attr_length: uint16 {
 		self.value_padding = padding($$);
-		self.total_size = $$ + self.value_padding;
+		self.total_size = cast<uint32>($$) + self.value_padding;
```
bbannier, line 166: This is a good solution, but a little subtle. Maybe instead:
```suggestion
		self.total_size = cast<uint32>($$) + self.value_padding; # Store total size as `uint32` so this operation cannot overflow.
```

bbannier (changes requested):

analyzer/stun/stun.spicy
awelzel: Thanks!

tests/protocol/stun/attribute-length-ffff.zeek
awelzel: I've added a `zeek-cut uid name addl` step - this should work on either Zeek version.

bbannier (approved):
Thanks @awelzel!

PR #111, opened 2021-12-06
Move PE analyzer to zeek/spicy-pe.

bbannier:


zkg.meta
```diff
@@ -11,6 +11,7 @@ test_command  = unset -v CXX CXXFLAGS LD LDFLAGS && cd tests && PATH=$(zkg confi
 # express that dependency due to zeek/package-manager#106
 # depends = spicy-plugin >=1.2.2
 depends = http://github.com/zeek/spicy-ldap ==0.0.2
+    http://github.com/zeek/spicy-pe ==0.0.1
```
rsmmr, line 14: Question: Do we really want to pin versions? I get why that's desirable from our perspective, but I'm wondering if that's maintainable: with so many external analyzers, it'll be regularly out of date.

rsmmr (approved):

zkg.meta
bbannier: I opened #112 for moving to unpinned deps and will update this PR once that one is merged.

bbannier:
Converting this to draft since I'll add changes to move all analyzers to separate packages.

bbannier:
The PR to add these analyzer packages to the package index is zeek/packages#171.

rsmmr (approved):

PR #118, opened 2022-07-07
Move CI to upstream Zeek images.

bbannier:
We also extend the test matrix to test both against the latest release
and the current development snapshot.


bbannier:
Closing to retrigger with current images.

PR #122, opened 2024-03-05
Drop spicy-ldap from dependencies.

bbannier:
As of zeek-6.1 spicy-ldap is part of Zeek. Drop if from the list of dependencies. Users who still want to install this package should install it explicitly.


bbannier:
The failing tests on `zeek-dev` are due to validation changes in Spicy, see zeek/spicy-dhcp#13 or zeek/spicy-ldap#30 (not required with this PR) for the cleanups.

awelzel (approved):

